PR #1, opened 2019-10-11
Fixed both typos on line 95

kant:


bwstefano:
Thanks for your review, but this is under a very initial stage of development, so naturally typos and other issues will appear. Therefore I'm not accepting this request for now.

PR #2, opened 2019-11-14
Use lando to setup development environment.

barraponto:
This new setup makes development setup as easy as one command (see README.md).
The only change for non-lando developers is in the build step: we now just npm run build and copy files (or symlink) over to wp-content/plugins.


barraponto:
@TODO:

[ ] Document installing phpunit with composer and running it with `./vendor/bin/phpunit` for non-lando setups
[ ] Document rsync `./src/js/src` folder to `/path/to/wordpress/wp-content/plugins/jeo-plugin`

PR #199, opened 2020-10-23
Story Map fixes - responsive

pamelars86:


pamelars86:
@hockpond 
Please, could you test :
- Create a post with JEO Map block and view the published post
- Create a post with JEO One time map block  and view the published post
- Create a map, publish it and  and view the published map (link with format localhost/maps/name-map

In order to verify that your change
```
div.jeomap {
	position: relative; /*OLD*/
	position: absolute; /*NEW*/
}
```
don't impact them

PR #276, opened 2021-01-06
Bugfix/storymap 274

hockpond:


pamelars86:
@hockpond 
2 changes:

**1. Add same logic in "current slide":** 
- If slide doesn't have a title, current slide will show the number of the slide
- If slide has a title, current slide will show that title.

**2. Story map is not using font in jeo settings**
image

image

I used those fonts here, maybe they could help
image

PR #277, opened 2021-01-07
Feature/sidebar url

pamelars86:


pamelars86:
1. URL in email is not complete
image


2. Discovery embed: Mapbox is adding its own watermark and it is over site's logo:

image

PR #301, opened 2021-01-15
Bugfix/links inside layer content

isaquemelo:


isaquemelo:
Main test:
- Create / edit a map layer
- Add a link inside the paragraph using the "**Open in new tab**" option (print 1)
- Save the layer
- Create a storymap and save it
- Take a look at the front-end to see if the JSON was parsed correctly 

Compatibility test:
- Open an old storymap created before theses changes, the console should show that the block was updated (print 2)

image
image

PR #309, opened 2021-01-27
Bugfix/force storymap update

isaquemelo:


pamelars86:
2 issues:
- When layer is deleted, story map is broken
- When layer's visibility changes from 'public' to 'private', story map is broken

PR #343, opened 2021-03-03
Add auto width to allow theme overwrite

isaquemelo:


pamelars86:
**Front End**

Map Block is working correctly 

image

PR #352, opened 2021-03-30
Feature/visibility control

isaquemelo:


pamelars86:
It should be related to
https://github.com/EarthJournalismNetwork/jeo-theme/issues/450

pamelars86:
Pending fix
- [ ] When I tried to open a Embed Url of a map with "hide embed URL", it should redirect to HomePage. Currently, that URL opens a blank page

image


If that URL Embed is used as an iframe in a Post, the Home will appear on it.

PR #358, opened 2021-03-31
Add editor settings to limit toolbar itens

isaquemelo:


pamelars86:
**Before**

image


**After**

image

PR #489, opened 2026-03-12
Migrate all blocks to Block API v2 for WP 7 compatibility

### LINKED ISSUE #487: Migrate blocks to Block API v2 for WordPress 7 iframe editor compatibility
## Problem

Since WordPress 6.9, the plugin's four blocks are generating deprecation warnings:

```
Block with API version 2 or lower is deprecated since version 6.9.
```

Affected blocks:
- `jeo/map` (registered with API version 1)
- `jeo/onetime-map` (registered with API version 1)
- `jeo/storymap` (registered with API version 1)
- `jeo/embedded-storymap` (registered with API version 1)

These warnings are preventing map and layer creation in WordPress 6.9+ and will become breaking in WordPress 7, where all editors are planned to work as iframes.

**Reference:** https://developer.wordpress.org/block-editor/reference-guides/block-api/block-api-versions/block-migration-for-iframe-editor-compatibility/

### Changes required

1. **Add `apiVersion: 2`** to all four `registerBlockType` calls
2. **Implement `useBlockProps()`** in all edit components (block wrapper identification)
3. **Implement `useBlockProps.save()`** in all save components (proper HTML output)
4. **Add `deprecated` migrations** for backward compatibility with existing saved content
5. **Update PHP `register_block_type`** calls with `api_version => 2`
6. **Update PHP render callbacks** to handle both legacy (raw JSON) and v2 (JSON wrapped in `<div>`) save formats

### Files affected

- `src/js/src/map-blocks/index.js` — block registrations, save functions, deprecated entries
- `src/js/src/map-blocks/map-editor.js` — edit component
- `src/js/src/map-blocks/map-display.js` — save component
- `src/js/src/map-blocks/onetime-map-editor.js` — edit component
- `src/js/src/map-blocks/onetime-map-display.js` — save component
- `src/js/src/map-blocks/storymap-editor.js` — edit component
- `src/js/src/map-blocks/embedded-story-map-editor.js` — edit component
- `src/includes/class-jeo.php` — PHP block registration + render callbacks

### Backward compatibility

- Existing blocks saved with v1 are automatically migrated via `deprecated` array when re-edited
- PHP `extract_json_from_block_content()` helper handles both old and new save formats in render callbacks
- Frontend selectors (`.jeomap`, `map_id_*`) are preserved in the new save output

## Phase 2 — Migrate to API v3 (future)

After Phase 1 is stable, a second phase will bump to `apiVersion: 3` with additional changes:
- Fix `document.body` direct manipulation in storymap-editor (use `ownerDocument` via refs)
- Test MapGL/MapLibreGL canvas rendering inside iframe
- Test CKEditor 5 compatibility inside iframe
- Test react-beautiful-dnd inside iframe
- Verify CSS loading in isolated iframe context

bwstefano:
## Summary

Resolves #487

Migrates all four JEO map blocks from Block API v1 to v2, eliminating the deprecation warnings introduced in WordPress 6.9+ and ensuring compatibility with WordPress 7, where all editors run inside iframes.

### Changes

- **`src/js/src/map-blocks/index.js`** — Added `apiVersion: 2` to all four block registrations (`jeo/map`, `jeo/onetime-map`, `jeo/storymap`, `jeo/embedded-storymap`). Added `deprecated` entries for backward compatibility with existing saved content.
- **`src/js/src/map-blocks/map-display.js`** — Added `MapSave` component using `useBlockProps.save()` for API v2 save output.
- **`src/js/src/map-blocks/onetime-map-display.js`** — Added `OnetimeMapSave` component using `useBlockProps.save()` for API v2 save output.
- **`src/js/src/map-blocks/map-editor.js`**, **`onetime-map-editor.js`**, **`storymap-editor.js`**, **`embedded-story-map-editor.js`** — Updated edit components to use `useBlockProps()`.
- **`src/includes/class-jeo.php`** — Added `api_version: 2` to PHP `register_block_type()` calls. Added `extract_json_from_block_content()` helper to handle both legacy (raw JSON string) and API v2 (JSON wrapped in a `<div>` from `useBlockProps.save()`) render callback formats.

## Tested

- Verify no Block API deprecation warnings appear in the WordPress editor (WP 6.9+)
- Confirm existing maps saved with API v1 still render correctly (backward compatibility via `deprecated` entries)
- Confirm new maps created with the updated blocks save and render correctly
- Test all four block types: `jeo/map`, `jeo/onetime-map`, `jeo/storymap`, `jeo/embedded-storymap`
- Test storymap and embedded-storymap dynamic render callbacks with both legacy and v2 saved content

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/includes/class-jeo.php
+++ b/src/includes/class-jeo.php
@@ -430,7 +430,9 @@ class Jeo {
 			}
 		);
 
-		return '<div class="story-map-container" data-properties="' . htmlentities( wp_json_encode( $saved_data ) ) . '" ></div>';
+		$wrapper_attributes = get_block_wrapper_attributes( array( 'class' => 'story-map-container' ) );
+
+		return '<div ' . $wrapper_attributes . ' data-properties="' . htmlentities( wp_json_encode( $saved_data ) ) . '" ></div>';
 	}
 
 	public function filter_rest_query_by_zone( $args, $request ) {
--- a/src/includes/class-jeo.php
+++ b/src/includes/class-jeo.php
@@ -289,6 +289,7 @@ class Jeo {
 		register_block_type(
 			'jeo/storymap',
 			array(
+				'api_version'     => 2,
 				'render_callback' => array( $this, 'story_map_dynamic_render_callback' ),
 				'editor_script'   => 'jeo-map-blocks',
 				'editor_style'    => 'jeo-map-blocks',
@@ -297,6 +298,7 @@ class Jeo {
 		register_block_type(
 			'jeo/embedded-storymap',
 			array(
+				'api_version'     => 2,
 				'render_callback' => array( $this, 'embedded_story_map_dynamic_render_callback' ),
 				'editor_script'   => 'jeo-map-blocks',
 				'editor_style'    => 'jeo-map-blocks',
@@ -304,8 +306,34 @@ class Jeo {
 		);
 	}
 
+	/**
+	 * Extract JSON content from block save output.
+	 * Handles both legacy format (raw JSON string) and API v2+ format
+	 * (JSON wrapped in a <div> element with useBlockProps).
+	 *
+	 * @param string $content Block save output.
+	 * @return string Extracted JSON string.
+	 */
+	private function extract_json_from_block_content( $content ) {
+		$content = trim( $content );
+
+		// If it's already valid JSON, return as-is (legacy format)
+		$decoded = json_decode( $content );
+		if ( json_last_error() === JSON_ERROR_NONE ) {
+			return $content;
+		}
+
+		// API v2+ format: JSON is wrapped in a <div> element
+		// Strip the outer HTML wrapper to get the JSON content inside
+		$inner = preg_replace( '/^<div[^>]*>/', '', $content );
+		$inner = preg_replace( '/<\/div>$/', '', $inner );
+		$inner = trim( $inner );
+
+		return $inner;
+	}
+
 	public function embedded_story_map_dynamic_render_callback( $block_attributes, $content ) {
-		$content = json_decode( $content );
+		$content = json_decode( $this->extract_json_from_block_content( $content ) );
 
 		$story_id                       = $content->attributes->storyID;
 		$story                          = get_post( $story_id );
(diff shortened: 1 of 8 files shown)


src/includes/class-jeo.php
```diff
@@ -297,15 +298,42 @@ public function register_block_types() {
...
+		$inner = preg_replace( '/^<div[^>]*>/', '', $content );
+		$inner = preg_replace( '/<\/div>$/', '', $inner );
+		$inner = trim( $inner );
+
+		return $inner;
+	}
+
 	public function embedded_story_map_dynamic_render_callback( $block_attributes, $content ) {
-		$content = json_decode( $content );
+		$content = json_decode( $this->extract_json_from_block_content( $content ) );
```
leopiccionia, line 336: I don't understand the changes in storymap blocks ("save" mode).

The idea of adding a wrapping `<div>` element is to save block props (e.g. [additional CSS classes](https://wordpress.com/support/wordpress-editor/adding-additional-css-classes-to-blocks/)). In practice, though, all extra attributes are discarded. The old saving method would have identical results.

The block props seems to be used accordingly in story maps editor-side, and on map blocks, though.

bwstefano: I've asked Codex to evaluate what Claude did here (🤖🎉) based on your comment:

> The intention behind the wrapper in `save` was to move the storymap blocks to the API v2 pattern and keep the serialization aligned with `useBlockProps.save()`, while preserving compatibility with previously saved raw JSON.
> 
> That said, you're right that the current implementation stops halfway. `save()` now serializes the storymap JSON inside the wrapper from `useBlockProps.save()`, but `extract_json_from_block_content()` strips that wrapper and `story_map_dynamic_render_callback()` replaces it with a hard-coded `<div class="story-map-container">`. So in practice, custom classes and other wrapper attributes are discarded, and the old save shape is effectively equivalent for front-end output.
> 
> The way to make this change meaningful is to keep the wrapper in `save()`, but stop treating it as disposable in the dynamic render path: PHP should extract only the inner JSON payload from `$content` and then render the outer element with `get_block_wrapper_attributes( [ 'class' ='story-map-container' ] )` (and do the same for the embedded storymap path). That would preserve the API v2 migration goal while also ensuring wrapper props actually survive to the final rendered markup.

leopiccionia: Fixed by [9f9b6b9](https://github.com/InfoAmazonia/jeo-plugin/pull/489/commits/9f9b6b968744298d0dfc129fec91a66bff8485ba).

leopiccionia (approved):

PR #497, opened 2026-03-16
Migrate all blocks to Block API v3 for WordPress 7 iframe editor compatibility

### LINKED ISSUE #488: Migrate blocks to Block API v3 for WordPress 7 iframe editor
## Context

Follow-up to #487 (Block API v2 migration).

With `apiVersion: 3`, blocks are declared compatible with the WordPress iframe editor. This is required for WordPress 7, where all editors will run inside iframes.

## Key difference from v2

In the iframe editor, the block's edit component runs inside an iframe. This means:
- `document` and `window` refer to the iframe's document/window, not the parent admin page
- Scripts injected via `wp_localize_script` are available inside the iframe
- CSS must be properly enqueued as editor styles to load inside the iframe
- Third-party libraries that attach listeners to `document`/`window` globals may need adjustments

### Step 1 — Bump `apiVersion` to 3
Minimal change: update all four blocks from `apiVersion: 2` to `apiVersion: 3` in both JS and PHP registrations.

### Step 2 — Fix `document.body` direct manipulation
`storymap-editor.js` sets a CSS custom property directly on `document.body`. In the iframe context, this needs to use `ownerDocument` from a ref to target the correct document.

### Step 3 — Test third-party libraries in iframe
Libraries that may have issues:
- **CKEditor 5** — heavy DOM manipulation, may attach listeners to wrong document
- **react-beautiful-dnd** — uses global `window`/`document` event listeners
- **MapLibreGL / Mapbox GL** — canvas rendering (likely works, needs verification)

## Depends on

- #487 (Block API v2 migration)

bwstefano:
Supersedes #490.

This replacement PR preserves the original commit organization and branch stacking, and only normalizes the Codex co-author trailers after the original PR was closed by a history rewrite.

---

## Summary

Resolves #488

This PR migrates all JEO blocks from Block API v2 to v3, ensuring full compatibility with WordPress 7's iframe-based editor. The migration required a series of targeted fixes to address the cross-document isolation introduced by the iframe context.

### What changed

- **Block API v3 bump** — All four blocks (`map`, `map-layer`, `storymap`, `onetime-map`) updated to `apiVersion: 3` in both JS and PHP registrations
- **Cross-document `instanceof` patch** — Added a guard to make `HTMLElement instanceof` checks work across the iframe/parent document boundary, preserving subclass checks
- **`document`/`window` scoping fixes** — Replaced bare `document.body` and global `document` references with `ownerDocument`-aware equivalents throughout `storymap-editor.js` and MapGL initialization
- **Sidebar portals** — Fixed React error 200 caused by sidebar plugins not mounting inside the iframe; panels now render inline in the post content area using `PluginDocumentSettingPanel`
- **Map/layer preview blocks** — Added dedicated preview blocks rendered in the post content area so authors can see a live preview without relying on sidebar portals
- **`crypto.randomUUID` polyfill** — Added a fallback for environments where the native API is unavailable inside the iframe
- **MapLibreGL container error** — Fixed the GL context initialization crash that occurred when the map container was not yet attached to the iframe document
- **Storymap settings & `getStyleUrl`** — Fixed a crash in `onetime-map` caused by calling `getStyleUrl` before map data was available; restored storymap settings panel in the iframe editor
- **Drag interaction** — Restored map drag/pan behavior after the iframe migration broke pointer event forwarding
- **SVG icons** — Replaced deprecated Dashicons with `@wordpress/icons` SVG equivalents in the storymap editor toolbar
- **CKEditor upgrade** — Upgraded the storymap rich-text editor to CKEditor 5 and refined editor lifecycle behavior for the iframe context

## Tested
- Open a post/page in the WordPress block editor (WP 7 / Gutenberg with iframe editor enabled)
- Insert a **Map** block — verify it renders and is interactive (pan, zoom, drag)
- Insert a **Map Layer** block — verify preview appears in the content area
- Insert an **Onetime Map** block — verify no JS crash on load
- Open a **Storymap** post — verify CKEditor loads, toolbar icons appear, and settings panel is accessible
- Confirm the sidebar shows map/layer previews via `PluginDocumentSettingPanel`
- Publish a post containing a map block — verify the publish button works without error

This branch was built on top of the work from https://github.com/InfoAmazonia/jeo-plugin/pull/489 and therefore includes those commits as part of the current diff.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/js/src/lib/maplibregl/maplibregl-loader.js
```diff
@@ -4,6 +4,147 @@ import { isMapboxURL, transformMapboxStyle, transformMapboxUrl } from 'maplibreg
 
 import 'maplibre-gl/dist/maplibre-gl.css'
 
+// Fix: Patch instanceof checks for cross-document (iframe) compatibility.
```
leopiccionia, line 7: Sou cético de que esse código tenha alguma utilidade:
1. A função `_resolveContainer` tem suporte para iframes, tanto que já verifica se `nodeType === 1` (ver [código](https://github.com/maplibre/maplibre-gl-js/blob/a8cf500b71089778f93c64c975bcaa0598e14c5d/src/ui/map.ts#L3384)), que é o patch proposto
2. A função `isPointableEvent` **não** compara com as globais do documento atual; ela explicitamente compara com as globais do `ownerDocument` da elemento `target` do evento (ver [código](https://github.com/maplibre/maplibre-gl-js/blob/a8cf500b71089778f93c64c975bcaa0598e14c5d/src/util/util.ts#L1174)), usando as globais do documento atual apenas para eventos sem `target` (que não é o caso)
3. Eventos não se propagam (_bubble up_) entre documentos (por questões óbvias de privacidade), mesmo se eles possuem o mesmo domínio

Remover esses patches causa algum problema?

src/js/src/util/generate-uuid.js
leopiccionia, line 1: I don't like this change, at all:

1. `crypto.randomUUID` is [widely available since 2022](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID).
2. Do we really need (and should) support insecure (non-HTTPS, non-localhost) contexts?
3. `crypto.getRandomValues` is widely available since 2015. We don't need to go further.
4. I'd rather use the brilliant [`@lukeed/uuid`](https://www.npmjs.com/package/@lukeed/uuid) than roll my own crypto.

src/js/src/layers-sidebar/layer-preview-portal.js
```diff
@@ -1,6 +1,13 @@
 import { createPortal } from '@wordpress/element';
 
 export default function LayerPreviewPortal( { children } ) {
+	// Only look in the parent document. In Block API v3 the editor
+	// content lives inside an iframe — createPortal cannot reliably
+	// render across document boundaries, so we fall back to inline
+	// rendering (the preview shows in the sidebar panel instead).
```
leopiccionia, line 7: Does this fallback behavior make sense?

src/includes/class-jeo.php
```diff
@@ -702,6 +751,20 @@ public function restrict_story_map_block_count() {
 				array( 'jeo/storymap' ),
 			);
 			$wp_post_types['storymap']->template_lock = 'all';
+
+			if ( isset( $wp_post_types['map'] ) ) {
+				$wp_post_types['map']->template      = array(
+					array( 'jeo/map-editor', array( 'align' => 'full' ) ),
+				);
+				$wp_post_types['map']->template_lock = 'all';
+			}
```
leopiccionia, line 766: I think locking this template is a breaking change. I'm not sure if any active JEO instance uses the Gutenberg editor to enhance the map single, but this is supposedly supported.

If so, the block injection via PHP should be updated accordingly, to prepend the previewer block instead of replacing existing content ([locking the removal/moving of the previewer block](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-templates/#individual-block-locking) is a good idea, though).

src/js/src/map-blocks/map-editor-preview.js
```diff
@@ -0,0 +1,173 @@
...
+	const setPostMeta = useCallback( ( meta ) => editPost( { meta } ), [ editPost ] );
+
+	const [ zoomState, setZoomState ] = useState( 'initial_zoom' );
+	const [ key, setKey ] = useState( 0 );
+	const mapRef = useRef( undefined );
+
+	// Expose setPanLimitsFromMap for the sidebar MapSettings button.
+	// Since the block runs inside the iframe and the sidebar in the parent,
+	// we use the editor data store to bridge them: MapSettings dispatches
+	// a custom action, and this block listens via a window message (future).
```
leopiccionia, line 34: It seems like this `postMessage` has not been implemented. Therefore, the values displayed on block panel may get out of sync.

src/js/src/maps-sidebar/map-preview-portal.js
```diff
@@ -1,6 +1,13 @@
 import { createPortal } from '@wordpress/element';
 
 export default function MapPreviewPortal( { children } ) {
+	// Only look in the parent document. In Block API v3 the editor
+	// content lives inside an iframe — createPortal cannot reliably
+	// render across document boundaries, so we fall back to inline
+	// rendering (the preview shows in the sidebar panel instead).
```
leopiccionia, line 7: Same question as in layer preview portal: does this fallback behavior make sense?

src/js/src/lib/maplibregl/maplibregl-loader.js
bwstefano: Faz sentido. Parece que a única parte que ainda parece tratar de um problema separado é `FullscreenControl`, que no código upstream ainda depende de `window.document` e de `instanceof HTMLElement`. Então talvez essa parte ainda deveria ser considerada.

src/js/src/util/generate-uuid.js
bwstefano: Agreed.

src/js/src/layers-sidebar/layer-preview-portal.js
bwstefano: After reviewing the final head of #497, this fallback does not appear to be part of the active preview path anymore.

During the fixes under this PR, the layers' preview was moved to the sidebar, but later moved back into the editor content area through the `jeo/layer-editor` block, and Codex could not find any remaining import/use sites for `LayerPreviewPortal` in the final branch head. So I agree the fallback and the comment in this file are misleading. This file should either be cleaned up or removed.

src/includes/class-jeo.php
bwstefano: `inject_editor_block_for_map()` / `inject_editor_block_for_layer()` replaces `content.raw` with only the preview block when it is missing, and `map` / `map-layer` are still set to `template_lock = 'all'`. Since those post types still support the core editor, this can hide or discard existing body content instead of prepending the preview block.

So I agree this is a legitimate compatibility concern.

src/js/src/map-blocks/map-editor-preview.js
bwstefano: The comment in this file talked about a future cross-frame bridge, but #497 did not actually implement a `postMessage` bridge there. So, as of #497, the “Set current as map settings” path was indeed left in an incomplete state.

This was fixed later in stacked PR #503. The follow-up there added a `handleSetPanLimitsFromMap()` bridge in `MapPanel` that first uses the local callback, then `window.__jeoSetPanLimitsFromMap`, and finally resolves the iframe window and calls the function there. So the latest stacked branch no longer has that broken flow.

src/js/src/maps-sidebar/map-preview-portal.js
bwstefano: Same conclusion as the layer preview portal thread.

By the final head of #497, map preview rendering had already been moved back into the editor content area through the `jeo/map-editor` block, so this portal file is stale code rather than part of the active preview flow.

leopiccionia: I'll resolve them, now, and do any cleanup, if still needed, at the end of the queued PRs.

src/includes/class-jeo.php
leopiccionia: Fixed by 2724900 and 237b3d8.

src/js/src/lib/maplibregl/maplibregl-loader.js
leopiccionia: Em relação ao `FullscreenControl`, precisamos testar esses blocos no 7.0 para verificar se o comentário do patch está correto.

Se o Mapbox GL/MapLibreGL usa o `document` de fora do iframe, o comentário e o patch estão corretos. Caso contrário, o fullscreen causaria erro, pois o iframe do Gutenberg, à primeira vista, não adiciona o [`allowfullscreen`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLIFrameElement/allowFullscreen) ao HTML do iframe.

src/js/src/util/generate-uuid.js
leopiccionia: Do you prefer to revert to native `crypto.randomUUID` or to use the `@lukeed/uuid` library?

src/js/src/lib/maplibregl/maplibregl-loader.js
bwstefano: Está bem. Eu tinha testado e funcionava, mas não sei se já num beta do 7.0 ou no 6.9.4 ainda. Nem tenho certeza se já neste PR ou depois do último PR.

src/js/src/util/generate-uuid.js
bwstefano: I think we could revert to crypto.randomUUID(), which solves our issue easily and don't add any newer dependency.

leopiccionia: Fixed by https://github.com/InfoAmazonia/jeo-plugin/pull/497/commits/311076c2dfecb6f7d69d3764fa64110864ad5e96.

src/js/src/lib/maplibregl/maplibregl-loader.js
leopiccionia: Nos meus testes, o `FullScreenControl` funcionou no editor com iframe do WordPress 7.

PR #504, opened 2026-03-16
Fix baseline map and storymap regressions

bwstefano:
## Summary
- fix published-map related post handling for null coordinates and dense same-point clusters
- replace brittle same-point spiderfy timing with a navigable popup flow
- stabilize storymap slide ordering with explicit controls and focused regression coverage
- update the dependency upgrade reports to close Batch 1

## Testing
- npm run test:unit
- npm run build
- php scripts/check-php-compat.php


src/js/src/jeo-map/class-jeo-map.js
```diff
@@ -665,6 +682,23 @@ export default class JeoMap {
...
+							map.addLayer( {
+								id: 'unclustered-points-hitarea',
+								type: 'symbol',
+								source: 'storiesSource',
+								filter: [ '!', [ 'has', 'point_count' ] ],
+								layout: {
+									'icon-image': 'news-marker',
+									'icon-size': parseFloat(
+										jeoMapVars.images['/js/src/icons/news-marker'].icon_size
+									) * 1.6,
```
leopiccionia, line 694: I'll need to double-check why it's multiplying by `1.6` and figure out if this causes any image distortion.

leopiccionia (approved):

src/js/src/jeo-map/class-jeo-map.js
bwstefano: It’s multiplied by `1.6` only to enlarge the invisible `unclustered-points-hitarea` click target, so overlapping related-post markers are easier to hover/click. It was not added to change the visible marker size.

src/js/src/jeo-map/class-jeo-map.js
```diff
@@ -727,12 +775,18 @@ export default class JeoMap {
 								},
 							} );
 
-							this.spiderifier.applyTo('unclustered-points');
```
leopiccionia: É esperado a remoção da ativação do spiderifier aqui?

Aparentemente não é mais possível ver o popup de detalhes de um post clusterizado, só o popup que mostra o número total de pontos clusterizados.

PR #507, opened 2026-03-17
Align frontend runtime tracking

bwstefano:
## Summary
- add a local react-autosize-textarea override so stale peer metadata stops polluting npm installs
- record the resolved React, ReactDOM and Gutenberg editor package line in the dependency report baseline
- document that react-leaflet 5 remains blocked because the published Gutenberg editor stack still peers on React 18

## Checks
- npm ci
- npm ls react-autosize-textarea react react-dom
- npm run test:unit
- npm run build
- php scripts/check-php-compat.php


leopiccionia:
Although the `react-autosize-textarea` warning is annoying, do we really need to vendor a package we probably don't even use?

This dependency is not even bundled, as WordPress/Gutenberg runtime code is removed thanks to `@wordpress/dependency-extraction-webpack-plugin`.

bwstefano:
I agree, @leopiccionia.

This PR was created as part of an effort intended to update React to version 19, after which I realized Gutenberg is still on React 18. Therefore, `react-leaflet` 5 remains blocked. But given that, I agree vendoring `react-autosize-textarea` is probably heavier than necessary.

I shouldn't have kept it as a PR and I’m fine with dropping the override and keeping only the conclusion/documentation from the investigation.

leopiccionia:
> I’m fine with dropping the override and keeping only the conclusion/documentation from the investigation.

I'm fine with it, too.

PR #508, opened 2026-03-17
chore: extend PHP 8.5 compatibility coverage

bwstefano:
## Summary
- extend the repository-owned PHP compatibility review to PHP 8.5
- add experimental PHP 8.5 coverage to the PHP compatibility and WordPress smoke workflows
- document that WordPress core 6.9 treats PHP 8.5 as beta support while stable WP-CLI support is still planned upstream

## Testing
- composer validate --no-check-publish
- php scripts/check-php-compat.php on PHP 8.0, 8.1, 8.2, 8.3, 8.4 and 8.5
- WP_CLI_PHP=/opt/homebrew/bin/php WP_DIR=/tmp/jeo-plugin-wordpress-smoke-85 bash scripts/wordpress-smoke.sh


README.md
```diff
@@ -30,6 +37,8 @@ npm ci
 npm run build
 npm run test:unit
 php scripts/check-php-compat.php
+/opt/homebrew/opt/php@8.5/bin/php scripts/check-php-compat.php
```
leopiccionia, line 40: We don't list Homebrew as a requirement, do we? It's not usually installed by default in Linux distros.

leopiccionia (approved):

README.md
bwstefano: This was probably introduced because `/opt/homebrew/opt/php@8.5/bin/php` is the path for my local PHP 8.5 setup, and that ended up being copied into the README as if it were a general project instruction. It should not be documented that way, because Homebrew is not a repository requirement and those commands should be written in a portable form, with the Homebrew path mentioned only as an optional local example if needed.

bwstefano:
```suggestion
php scripts/check-php-compat.php
```

leopiccionia: I don't know a portable way to run a specific PHP version. If you had the `php8.5` binary in your environment, something like `/usr/bin/env php8.5 scripts/check-php-compat.php` should work on most POSIX environments.

If any `php` binary is used, no matter the aliased version,  `/usr/bin/env php scripts/check-php-compat.php` would suffice.

PR #511, opened 2026-03-18
Follow up sanity fixes for editor maps, geocoding, and Discovery

bwstefano:
## Summary

This follow-up branch closes the baseline issues that were left out of the main sanity track and packages the work into four areas:

### 1. Editor sidebar and map runtime follow-ups
- keep the map, layer, and storymap editors on the Document sidebar instead of bouncing back to Block on selection
- load the map runtime CSS in the block-editor iframe and related sidebar contexts so the Mapbox GL missing-CSS warning stops appearing in the editor
- normalize the hidden and expanded Storymap slides settings panel widths

### 2. Post geocoding reliability
- stabilize the geocoding field state after selecting an address, dragging a marker, canceling, saving, and reopening
- keep the current address visible and actionable so new and existing points can be saved after drag or search updates
- add an accent-insensitive fallback for address search so queries such as `São Paulo`, `Belém`, and `Amapá` work without forcing users to strip diacritics manually

### 3. Discovery markers, hover sync, sharing, and filters
- switch Discovery GeoJSON features to stable per-point IDs while preserving the story/post ID in feature properties
- keep marker, cluster, `View in map`, and card hover states in sync across multi-point stories and clustered results
- scroll the stories list to the hovered marker or cluster and reflect list hover back onto markers or clusters in the map
- increase the initial Discovery fetch window so more geolocated posts appear before filters are applied
- replace Twitter and Facebook sharing with X and WhatsApp
- include active story filters and selected layers in shared Discovery URLs and restore that state when opening the shared link
- fix the tag-reset flow so removing a selected tag returns Discovery to the default result set
- restore normal layer application in Discovery after the shared-layer initialization changes

### 4. Documentation close-out
- remove the resolved baseline follow-ups from the upgrade report and smoke checklist
- close the stale override-governance note
- reclassify `react-leaflet` 5, PHPCS 4, and `wp-cli` PHP 8.5 noise as external monitoring paths instead of pending product bugs

## Notes
- no REST or post-meta contract changes are introduced in the Discovery work; only the internal feature ID strategy changes
- this branch is intended to merge into `codex/plugin-sanity-track`

## Testing
- `git diff --check`
- `php scripts/check-php-compat.php`
- `PATH="/opt/homebrew/opt/node@24/bin:$PATH" npm run test:unit`
- `PATH="/opt/homebrew/opt/node@24/bin:$PATH" npm run build`


src/js/src/shared/story-geojson.js
```diff
@@ -4,6 +4,22 @@ function toFiniteCoordinate( value ) {
...
+function hashFeatureKey( value ) {
+	let hash = 2166136261;
+
+	for ( let index = 0; index < value.length; index++ ) {
+		hash ^= value.charCodeAt( index );
+		hash = Math.imul( hash, 16777619 );
+	}
+
+	return hash >>> 0;
+}
```
leopiccionia, line 16: It seems to be the 32-bit [Fowler-Noll-Vo hash function](https://en.wikipedia.org/wiki/Fowler%E2%80%93Noll%E2%80%93Vo_hash_function).

leopiccionia (approved):

PR #516, opened 2026-03-23
Fix WordPress.org compliance issues

bwstefano:
## Summary
- document third-party and external services for WordPress.org review
- harden escaping, sanitization, embeds, and package metadata for the distributed plugin
- restore editor, storymap, legend, and Discovery behavior that regressed during the compliance pass
- add final stability fixes for Discovery REST payloads and complete the validation pass for the stacked PR
- finish responsive map, Related Posts, and storymap interaction polish for the published frontend and editor

## Context
This PR is intentionally stacked on top of `codex/i18n-string-review` and should be reviewed after that PR.

The goal here is to address the WordPress.org review concerns around:
- undocumented use of third-party or external services
- variables and options that must be escaped when echoed
- avoiding direct cURL usage and preventing regressions in transport/compliance rules

### Third-party services and privacy
- added a dedicated third-party services section to the plugin readme
- added privacy policy content for services used by the plugin
- clarified admin-facing disclosures for Mapbox and Nominatim usage

### Remote dependencies and compliance hardening
- removed avoidable remote glyph and marker asset dependencies
- replaced editor marker assets with local plugin assets
- kept user-configurable typography and logo URLs while documenting and sanitizing them appropriately
- added automated compliance checks for banned patterns and required documentation
- added a post-build compliance patch step for generated assets
- aligned plugin headers, direct access guards, and readme metadata with WordPress.org expectations

### Escaping, sanitization, and rendering
- tightened contextual escaping in settings, embeds, and map rendering
- normalized settings sanitization for typed values and asset URLs
- hardened embed output and automatic logo sizing for large images
- removed recursive map content rendering when outputting layer attribution/content on single maps

### Geocoding and editor behavior
- changed Nominatim lookups to explicit user-triggered search behavior
- added caching on the geocoder side
- fixed repeated search behavior without saving
- restored Enter-triggered address search
- fixed search field and button alignment in post geolocation and storymaps
- prevented premature “no results” messaging
- patched deprecated light API calls used by the map runtimes in editor flows

### Discovery, layers, and legends
- restored legend rendering for incomplete and legacy legend metadata paths
- isolated classic legend and layer type scripts to avoid collisions between shared globals
- preserved explicit map-layer ordering through a dedicated JEO REST route instead of filtered collection queries
- restored immediate layer reordering in editor previews and published maps
- reduced Discovery map collection responses to the fields it actually needs so map content is not rendered for every collection item
- stabilized Discovery layer loading, map expansion, and apply/remove flows
- kept the Discovery share action as `Embed` in Portuguese and Spanish, while preserving translated wording in other contexts
- updated new and edited strings in `pt_BR` and `es_CO`

### Runtime, Related Posts, and responsive layout polish
- stabilized Related Posts filter serialization, loading, and cluster behavior across Mapbox and MapLibre
- improved Discovery story pagination, scroll/hover coordination, and cumulative deduplication while keeping layer browsing responsive
- fixed storymap navigation edge cases, including the final navigation CTA behavior and the `Back to start` resize flow
- restored free movement in the layer editor preview instead of snapping back to the default map center and zoom
- aligned left and right map blocks more closely with the core Image block while preserving responsive behavior on smaller screens
- constrained map “more info” overlays to the available map area on smaller screens for both standard maps and one-time maps

## Validation
- `npm run check:env`
- `npm run build`
- `npm run build:report`
- `npm run test:unit`
- `vendor/bin/phpcs --standard=phpcs.xml.dist src`
- `php scripts/check-php-compat.php`
- WordPress smoke test on WordPress 6.9.4 with PHP 8.4
- WP Plugin Check in the smoke-test WordPress for the installed `jeo` plugin and a clean copied package; only the existing `meta_query` warnings remain
- local WordPress log review after exercising Discovery and core JEO flows; no new PHP or nginx log entries were added after the final fixes

## Notes
- local `wp-cli` still emits deprecation warnings during `i18n:json`, but the build completes successfully and the warnings come from the toolchain, not plugin runtime
- macOS can recreate `.DS_Store` in the local workspace, so the final Plugin Check confirmation was done against a clean copied plugin inside the smoke-test WordPress
- the `mapBlocks` bundle budget was updated to reflect the current stable CKEditor footprint used by the storymap editor


scripts/check-wporg-compliance.mjs
leopiccionia, line 1: Automated WordPress store checks are provided as a PHPCS plugin.

I've used [`wp-coding-standards/wpcs`](https://github.com/wordpress/wordpress-coding-standards) successfully in the recent past, in another project.

src/languages/jeo-pt_BR.po
```diff
@@ -512,6 +512,16 @@ msgstr ""
 "Quando selecionado, o Mapbox carrega os ativos oficiais do Mapbox GL JS de "
 "api.mapbox.com e usa os termos e práticas de privacidade do Mapbox."
 
+#: includes/settings/settings-page.php:93
+msgid ""
+"Visitors load Mapbox assets in the browser, and Mapbox receives the "
+"configured access token plus the visitor IP address and requested map "
+"resources."
+msgstr ""
+"Os visitantes carregam os ativos do Mapbox no navegador, e o Mapbox recebe "
```
leopiccionia, line 521: Não verifiquei todas as traduções, mas essa aqui me chamou a atenção. A tradução "assets" -> "ativos" faz mais sentido no contexto da Economia.

Traduziria, também, "the requested map resources" como "recursos do mapa solicitado".

src/js/src/shared/styles.js
```diff
@@ -1,7 +1,6 @@
 export const EMPTY_STYLE = {
 	version: 8,
 	name: 'Empty',
-	glyphs: 'https://fonts.openmaptiles.org/{fontstack}/{range}.pbf',
```
leopiccionia, line 4: Mapas e storymaps às vezes usam o `EMPTY_STYLE` como estilo base, quando nenhuma camada do Mapbox está configurada como estilo.

Não me lembro o contexto exato, mas essa adição melhorava a compatibilidade com o MapLibreGL nesses casos.

src/includes/maps/class-maps.php
```diff
@@ -701,14 +740,17 @@ public function add_capabilities() {
...
+			}
 
 			foreach ( $types as $type ) {
 				$role_obj->add_cap( "edit_{$type}" );
 				$role_obj->add_cap( "edit_{$type}s" );
 				$role_obj->add_cap( "delete_{$type}" );
 				$role_obj->add_cap( "read_{$type}" );
 				$role_obj->add_cap( "read_{$type}s" );
-				if ( 'contributor' !== $type ) {
+				if ( 'contributor' !== $role ) {
```
leopiccionia, line 753: Good catch!

src/includes/loaders.php
```diff
@@ -171,37 +175,145 @@ function jeo_register_embedder( $id, $base_url ) {
...
+
+/**
+ * Sanitize a font-family label for safe CSS output.
+ *
+ * @param string $font_name Raw font-family string.
+ * @return string
+ */
+function jeo_sanitize_font_family( $font_name ) {
+	$font_name = sanitize_text_field( (string) $font_name );
+	$font_name = preg_replace( '/[^A-Za-z0-9 \-_]/', '', $font_name );
```
leopiccionia, line 256: I don't understand why this sanitation is needed. There's nothing in [CSS spec](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/font-family#formal_syntax) that restricts the characters set of a font family name. I'd rather use generic HTML escaping.

src/js/src/discovery/blocks/stories-helpers.js
```diff
@@ -0,0 +1,27 @@
...
+		if ( Number.isFinite( storyId ) ) {
+			storiesById.set( storyId, story );
+		}
+	} );
+
+	return Array.from( storiesById.values() );
+}
+
+export function resolveStoriesPage( params = {}, pageInfo = {} ) {
+	if ( Object.prototype.hasOwnProperty.call( params, 'page' ) ) {
```
leopiccionia: [`Object.hasOwn`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn) is widely available and vastly superior.

src/js/src/lib/mapboxgl/mapboxgl-loader.js
```diff
@@ -123,6 +123,46 @@ function patchMissingCSSDetection( MapboxGL ) {
...
+		};
+	}
+
+	if ( typeof proto.setSky !== 'function' ) {
+		proto.setSky = function () {
+			return this;
+		};
+	}
+}
+
```
leopiccionia, line 165: We never formally supported it, so I'd rather find a workaround.

src/languages/jeo-pt_BR.po
bwstefano: Talvez trocar "assets" por "elementos"? "Ativos" também é um termo que me deixa encucado, mas não gosto de deixar "assets" no original. Pensei em "recursos", mas aí teria uma repetição do mesmo termo na frase.

Quanto a "requested map resources", acho que a tradução para "recursos do mapa solicitado" seria diferente do original. Para usar essa tradução, o original teria que ser "the requested map's resources", não?

src/js/src/discovery/blocks/stories-helpers.js
bwstefano: Agreed. Would you make the change? Should we also open an issue to review similar `hasOwnProperty.call(...)` usages across the JS codebase? That way, we could standardize them.

scripts/check-wporg-compliance.mjs
bwstefano: Agreed. I asked Codex to add the Node check as a narrow project safeguard, not as a replacement for the standard WordPress review tooling. I think the better direction is to rely on the PHPCS-based WordPress checks as the primary source of truth here, and keep this script only for project-specific assertions that the standard rulesets do not cover, such as our banned host patterns and the required Third Party Services section in the readme.

src/includes/loaders.php
bwstefano: Good point. The current regex is stricter than the CSS grammar and can strip valid font family names. I agree we should not enforce an artificial character whitelist here. However HTML escaping might not be the way to go because this value is ultimately composed into generated CSS, not just rendered as plain HTML text. Maybe doing minimal normalization on input and context-appropriate quoting/escaping when building the `font-family` declaration, so we preserve valid family names without over-sanitizing them?

src/js/src/shared/styles.js
bwstefano: Hum... O problema que fez a gente remover esse `glyphs` está [aqui](https://developer.wordpress.org/plugins/wordpress-org/common-issues/#calling-files-remotely):

> Offloading images, js, css, and other scripts to your servers or any remote service (like Google, MaxCDN, jQuery.com etc) is disallowed. When you call remote data you introduce an unnecessary dependency on another site. If the file you’re calling isn’t a part of WordPress Core, then you should include it -locally- in your plugin, not remotely.

Daí eles dizem pra gente incluir localmente quando possível e só em último caso vão permitir chamar uma fonte externa. Talvez a gente possa apelar por esse "último caso", então, [documentando apropriadamente](https://developer.wordpress.org/plugins/wordpress-org/common-issues/#undocumented-3rd-party) esta chamada ao OpenMapTiles.

O que você acha?

src/js/src/lib/mapboxgl/mapboxgl-loader.js
bwstefano: This did not come from the WordPress.org policy work itself, but from regressions that surfaced while validating that branch end to end.

The two patches were introduced for different runtime problems we hit during testing:

- `getLight`: we started seeing the deprecation warning `The map.getLight function is deprecated, prefer using map.getLights instead` in editor flows, in Console, especially when working with storymaps and slide transitions. The shim was added as a compatibility bridge so that code paths still expecting `getLight()` would stop warning while the rest of the branch remained testable.

- `getSky` / `setSky`: later, while testing the Mapbox runtime in the map editor, we hit a harder failure: `s.getSky is not a function`. At that point the preview could fail to initialize properly, so Codex added the no-op shim as a short-term workaround to keep the editor usable and allow the rest of the fixes in this PR to be validated.

A suggestion would be to identify the exact code paths still expecting `getLight()` / `getSky()` and make them runtime-aware instead of patching missing APIs into Mapbox. Does that make sense?

src/js/src/discovery/blocks/stories-helpers.js
leopiccionia: Fixed in [f47ca5a](https://github.com/InfoAmazonia/jeo-plugin/pull/516/commits/f47ca5ae6c4aa12d32d1a4ffddc6151816af3429).

My local [SonarQube](https://www.sonarsource.com/products/sonarqube/) instance (that's possibly not as powerful as CI-based one) flags `Object.prototype.hasOwnProperty.call(obj, key)` but not `obj.hasOwnProperty(key)`, but I replaced all occurrences on this branch for consistency.

PR #563, opened 2026-05-20
Fix public legend runtime UUID handling

bwstefano:
## Summary

This fixes a public map rendering failure triggered in local HTTP environments such as `infoamazonia.local`, where `crypto.randomUUID()` may be unavailable. In that context, the public `JeoLegend` bundle was creating UUIDs while normalizing barscale legend options, causing the map page to stop with `TypeError: crypto.randomUUID is not a function` before the map could open.

## Background

This follows the decision made in [PR #497](https://github.com/InfoAmazonia/jeo-plugin/pull/497), especially the discussion at [`discussion_r3001592940`](https://github.com/InfoAmazonia/jeo-plugin/pull/497#discussion_r3001592940):

- do not keep a custom `generate-uuid.js` helper;
- do not add a new `@lukeed/uuid` dependency;
- use native `crypto.randomUUID()` where UUIDs are actually needed;
- do not add fallback support for insecure, non-HTTPS/non-localhost contexts.

The problem here is slightly different from needing a fallback: the public map runtime should not need editor-only UUIDs at all. This PR respects the #497 agreement by removing UUID generation from the public legend normalization path instead of introducing a polyfill or dependency.

## What changed

- Removed `crypto.randomUUID()` usage from the public barscale defaults in `src/includes/legend-types/JeoLegend.js`.
- Updated barscale color normalization so it no longer creates missing `id` values at public render time.
- Preserved existing editor IDs during normalization when they are already present, so the editor can still keep stable list keys while working with legend options.
- Stripped barscale editor-only `id` values before saving normalized legend metadata.
- Replaced the remaining dangling `generateUUID()` calls in the legend editor code with native `crypto.randomUUID()`, matching the agreement from PR #497.

## Why this fixes the bug

The public map view only needs the legend data, such as colors and labels, to render the legend. The generated IDs are only needed by the React editor for list item identity while editing a legend. By keeping those IDs out of the public `JeoLegend` runtime, public maps no longer depend on `crypto.randomUUID()` and can render in the reported `http://infoamazonia.local/...` environment.

At the same time, editor code still uses native `crypto.randomUUID()` where new items are created, so this does not reintroduce the removed helper or add any dependency.

## Validation

- Applied the equivalent hotfix directly to the installed plugin under `infoamazonia.local` and opened:
  `http://infoamazonia.local/maps/alertas-de-desmatamento-deter-e-focos-de-calor-em-setembro-de-2022/`
- Confirmed the browser console no longer reported `crypto.randomUUID is not a function` or dangling `generateUUID` errors.
- Confirmed the map canvas was created and the public map opened in that local environment.
- Ran `npm run build` successfully. The build completed, with existing WP-CLI/PHP deprecation warnings emitted during i18n JSON generation.
- Ran `npm run test:unit -- --runInBand` successfully: 14 test suites passed, 43 tests passed.
- Ran `git diff --check` successfully.
- Checked the generated build output after build: public `js/build/JeoLegend.js` no longer contains `crypto.randomUUID()` or `generateUUID()`, and editor `js/build/layersSidebar.js` no longer contains `generateUUID()`.

## Notes

This intentionally does not add an insecure-context UUID fallback. The fix is to avoid UUID generation in the public map runtime and keep native UUID generation limited to editor-only code, as agreed in PR #497.

### DIFF
--- a/src/includes/legend-types/JeoLegend.js
+++ b/src/includes/legend-types/JeoLegend.js
@@ -5,6 +5,12 @@ function isPlainObject( value ) {
 	return Boolean( value ) && typeof value === 'object' && ! Array.isArray( value );
 }
 
+function omitEditorId( item ) {
+	const result = { ...item };
+	delete result.id;
+	return result;
+}
+
 function normalizeBarscaleColors( colors ) {
 	if ( ! Array.isArray( colors ) ) {
 		return [];
@@ -15,14 +21,12 @@ function normalizeBarscaleColors( colors ) {
 			if ( typeof color === 'string' ) {
 				return {
 					color,
-					id: generateUUID(),
 				};
 			}
 
 			if ( isPlainObject( color ) && typeof color.color === 'string' && color.color ) {
 				return {
 					...color,
-					id: color.id || generateUUID(),
 				};
 			}
 
@@ -185,14 +189,13 @@ class JeoLegend {
 			barscale: {
 				left_label: '0',
 				right_label: '100',
-				colors: [ {
-					color: '#ff0909',
-					id: crypto.randomUUID(),
+				colors: [
+					{
+						color: '#ff0909',
 					},
 					{
-					color:'#000',
-					id: crypto.randomUUID(),
-					}
+						color: '#000',
+					},
 				],
 			},
 		};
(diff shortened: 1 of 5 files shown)


bwstefano:
Added follow-up commit 0bb8535dae1d10dc259186f02cbe1453e2163ad2 to keep the legend editors aligned with the crypto approach from this PR.

What changed:

- Existing/default legend editor rows now receive deterministic editor-only IDs during mount (`barscale-color-${index}`, `simple-color-${index}`, `circle-${index}`, and `icon-${index}`) instead of calling `crypto.randomUUID()` while rendering existing metadata.
- Native `crypto.randomUUID()` remains in the add-new-row handlers only, so newly created legend rows still use the browser crypto API.
- The barscale editor no longer mirrors incoming props back into state with `getDerivedStateFromProps`, because that was replacing the constructor-normalized rows and dropping the IDs generated for editor state.

This avoids the unexpected editor crash when merely enabling or editing an existing legend in local/non-secure contexts, while preserving this PR's decision not to introduce a UUID fallback or polyfill for insecure origins.

PR #564, opened 2026-05-20
Restore storymap layer selections after editor hydration

bwstefano:
## Summary

This PR fixes a storymap editor regression where previously configured per-slide layer selections disappeared after upgrading to the current 3.0.0 release candidate.

Existing storymaps saved with JEO 2.14.1 still contained their `selectedLayers` data, and the selections would reappear after switching the plugin back to 2.14.1. The data was not actually gone initially; the 3.0.0 editor was failing to hydrate it safely and could then persist the empty selection state on save.

## Root cause

The storymap editor now hydrates map layers through the paginated `/jeo/v1/map-layer` REST helper instead of relying on an unbounded `getEntityRecords` request. That optimization made the editor initialize `loadedLayers` as an empty array while the layer request was still pending.

The existing reconciliation effect interpreted that initial empty array as a complete response, then filtered every slide's `selectedLayers` against it. Since no layer records were present yet, each slide was rewritten with `selectedLayers: []`.

There was a second compatibility risk: some older saved storymap payloads can carry layer IDs as strings while the newer map/layer records expose numeric IDs. Strict ID comparisons made those otherwise valid saved selections easier to drop during reconciliation and rendering.

## What changed

- Added a `hasResolved` signal to `useRecordsByIds()` so consumers can distinguish "request has not completed yet" from "request completed with no records".
- Updated the storymap editor to wait until the map-layer hydration request resolves before reconciling slide layer selections.
- Added a guard so a map that has configured layer IDs does not clear all slide selections merely because the layer hydration result is unexpectedly empty.
- Centralized storymap layer ID handling in tested helpers:
  - normalize layer IDs to positive integers
  - compare saved layer references across string/number ID formats
  - reconcile selected layers against both the parent map layers and hydrated map-layer records
- Updated the storymap editor's preview and layer toggle checks to use normalized ID matching.
- Updated the PHP storymap render callback to tolerate string IDs in saved storymap JSON while preserving the existing behavior of removing deleted, private, trashed, or detached layers.

## Decisions

- I kept reconciliation in place because it still protects users from stale references to deleted or removed layers. The fix is to run it only after layer data is actually available, not to remove the cleanup behavior.
- I normalized IDs at the helper/render boundary instead of changing the persisted storymap schema. This keeps backward compatibility for existing storymaps and avoids forcing a migration before users can safely reopen and save them.
- I did not include generated `src/js/build` assets in the commit because that directory is ignored in this repository. I did run the build locally to verify the editor bundle still compiles.
- I intentionally left the unrelated local `src/includes/storymap/class-storymap.php` change out of this PR so this branch only covers the storymap layer-selection regression.

## Validation

Automated checks run on the isolated PR worktree:

- `npm run test:unit` — passed, 14 suites / 46 tests
- `npm run build:assets` — passed, webpack compiled successfully and the compliance patch script completed
- `php -l src/includes/class-jeo.php` — passed
- `git diff --check` — passed

Focused regression coverage added:

- saved layer IDs represented as strings are normalized and matched against numeric map-layer IDs
- selected layers are sorted in parent map order without dropping valid legacy IDs
- reconciliation removes only layers that are absent from the parent map or the hydrated layer records

Local manual-test preparation:

- Applied the same fix directly to the JEO plugin installed at `infoamazonia.local`
- Copied the compiled `js/build` output into that local plugin installation
- Flushed the local WordPress cache with WP-CLI using a higher memory limit

Expected behavior after this fix: opening an existing storymap in the 3.0.0 editor should preserve the layers configured inside each slide instead of showing every slide as if no layers were selected.

### DIFF
--- a/src/includes/class-jeo.php
+++ b/src/includes/class-jeo.php
@@ -752,15 +752,30 @@ class Jeo {
 	 * @return bool
 	 */
 	private function layer_still_exists( array $map_layers, $selected_layer ) {
-		$layer_status = get_post_status( $selected_layer->id );
+		if ( ! is_object( $selected_layer ) || ! isset( $selected_layer->id ) ) {
+			return false;
+		}
+
+		$selected_layer_id = absint( $selected_layer->id );
+		if ( ! $selected_layer_id ) {
+			return false;
+		}
+
+		$layer_status = get_post_status( $selected_layer_id );
 		if ( 'trash' === $layer_status || false === $layer_status || 'private' === $layer_status ) {
 			return false;
 		}
 
 		foreach ( $map_layers as $layer ) {
-			if ( $layer['id'] === $selected_layer->id ) {
-				$selected_layer->meta->type               = get_post_meta( $layer['id'], 'type', true );
-				$selected_layer->meta->layer_type_options = (object) get_post_meta( $layer['id'], 'layer_type_options', true );
+			$map_layer_id = absint( $layer['id'] ?? 0 );
+			if ( $map_layer_id === $selected_layer_id ) {
+				if ( ! isset( $selected_layer->meta ) || ! is_object( $selected_layer->meta ) ) {
+					$selected_layer->meta = new \stdClass();
+				}
+
+				$selected_layer->id                       = $selected_layer_id;
+				$selected_layer->meta->type               = get_post_meta( $selected_layer_id, 'type', true );
+				$selected_layer->meta->layer_type_options = (object) get_post_meta( $selected_layer_id, 'layer_type_options', true );
 				return true;
 			}
 		}
@@ -830,7 +845,7 @@ class Jeo {
 
 			foreach ( $map_layers as $layer ) {
 				foreach ( $selected_layers as $selected_layer ) {
-					if ( $selected_layer->id === $layer['id'] ) {
+					if ( absint( $selected_layer->id ?? 0 ) === absint( $layer['id'] ?? 0 ) ) {
 						$selected_layers_order[] = $selected_layer;
 					}
 				}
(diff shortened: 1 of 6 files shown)


bwstefano:
Follow-up note from local validation: while checking the browser console output for an upgraded legacy storymap, Gutenberg logged `Updated Block: jeo/storymap` from WordPress core. That notice is not emitted by JEO; it appears because Gutenberg is normalizing older storymap content saved as raw JSON into the current block wrapper format generated by `useBlockProps.save()`.

The same console dump exposed that the editor-side storymap save cleanup removed `yoast_head` from layer objects but did not remove `yoast_head_json`. I added a second commit (`4ef1e3fb`) to keep the JavaScript save cleanup aligned with the PHP render cleanup by removing both fields from selected slide layers and navigation layers before serialization. The cleanup is also now defensive around incomplete legacy slide/layer objects.

Additional validation after this follow-up commit:

- `npm run test:unit -- storymap-ordering.test.js` passed, 8 tests
- `npm run build:assets` passed in the main worktree
- `npm run build:assets` passed in the isolated PR worktree
- `git diff --check` passed

Expected behavior: Gutenberg may still log a one-time block normalization notice for legacy storymaps until they are saved in the current wrapper format, but the rewritten storymap payload should no longer include the large `yoast_head_json` REST objects for every layer.

bwstefano:
Note about the extra commit `16b1b4e2` (`Keep legacy storymap Yoast payloads valid`):

The previous commit started removing `yoast_head_json` from saved storymap payloads, which is the desired behavior for new saves because it avoids persisting a large transient object from the REST API/Yoast. However, storymaps saved before that cleanup still include `yoast_head_json` in the block's serialized HTML.

Because Gutenberg validates blocks by comparing the saved HTML with the current `save()` output, those older storymaps would be reported as invalid. This extra commit keeps the cleanup in place for new saves, but adds a deprecated serializer that preserves `yoast_head_json` only to recognize the legacy markup and let the editor validate previously saved blocks without changing their stored content.

leopiccionia (approved):

PR #565, opened 2026-05-20
Harden storymap navigation mode

bwstefano:
## Summary

This PR hardens the storymap navigation mode in the frontend storymap display. It focuses on the plugin-side fragility we found while debugging the `Até a Última Gota - Colômbia` storymap locally.

It does not address the separate theme-side `notNavigatingMap` null access in `jeo-theme`; that error is still owned by the theme bundle.

## Problem

The storymap plugin had a few fragile DOM assumptions around the optional `Navigate the map` flow:

- navigation mode mixed React state with direct `style.display` mutations on `.navigate-map` and `.not-navigating-map`;
- the return-to-slides control was moved into `.navigate-map .jeomap`, which means Mapbox initialized with plugin markup inside the map container;
- fullscreen handling used an anonymous callback and queried `.return-to-slides-container` without checking whether the element exists;
- some lazy storymap setup paths assumed map/story DOM nodes were already present;
- the navigation button lived inside a `.step` with `opacity: 0.25`, so the whole button looked faded before the midpoint trigger activated it.

## Decisions

- Keep the return-to-slides control outside the Mapbox-owned `.jeomap` element. Mapbox expects to initialize an empty map container, so the storymap controls now stay as siblings inside `.navigate-map`.
- Use `isNavigating` state to switch between the story slides map and the navigable map, instead of manually mutating both containers in click handlers.
- Extract `enterNavigationMode()` and `exitNavigationMode()` so both return controls share the same exit behavior.
- Store fullscreen handling in `this.handleFullscreenChange` and remove it during unmount to avoid leaking event listeners.
- Guard DOM lookups with optional chaining where lazy initialization can run before a queried element exists.
- Apply full opacity to `.step.storymap-navigation-step`, not just the button itself, because the faded parent opacity was dimming the entire navigation step.

## Testing

- Built production assets successfully with Webpack and the WordPress.org compliance patch script.
- Ran the JS unit suite: `14` suites and `43` tests passed.
- Verified the local storymap at `http://infoamazonia.local/storymap/ate-a-ultima-gota-colombia/` with Playwright:
  - initial `Navegar no mapa` button opacity is `1`;
  - `.storymap-navigation-step` opacity is `0.99`;
  - before click, `.navigate-map` is hidden and `.not-navigating-map` is visible;
  - after clicking the navigation button and waiting for React state to apply, `.navigate-map` is visible and `.not-navigating-map` is hidden;
  - `.return-to-slides-container` remains a child of `.navigate-map`, while `.navigate-map .jeomap` only contains Mapbox/map controls;
  - browser console reported no warnings from this flow. The remaining console error is the known theme-side `app.js:839` null access, outside this PR.

### DIFF
--- a/src/js/src/jeo-storymap/storymap-display.js
+++ b/src/js/src/jeo-storymap/storymap-display.js
@@ -59,6 +59,15 @@ class StoryMapDisplay extends Component {
 		this.isIntroductionScrollLocked = false;
 		this.previousBodyOverflow = null;
 		this.previousDocumentOverflow = null;
+		this.handleFullscreenChange = () => {
+			const returnToSlidesContainer = this.el?.querySelector( '.return-to-slides-container' );
+
+			if ( returnToSlidesContainer ) {
+				returnToSlidesContainer.style.display = document.fullscreenElement ? 'none' : 'block';
+			}
+
+			window.scrollTo( 0, document.body.scrollHeight );
+		};
 
 		const slides = [];
 		props.slides.map( ( slide, index ) => {
@@ -137,6 +146,7 @@ class StoryMapDisplay extends Component {
 	componentWillUnmount() {
 		this.setIntroductionScrollLocked( false );
 		window.removeEventListener( 'resize', this.scroller.resize );
+		document.removeEventListener( 'fullscreenchange', this.handleFullscreenChange );
 	}
 
 	setIntroductionScrollLocked( locked ) {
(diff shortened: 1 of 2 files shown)


bwstefano:
I pushed `0dc383db` (`Prepare storymap navigation state for post embeds`) to make this PR own the shared storymap navigation/intro state that was previously overlapping with #567.

What this commit does:

- Adds `hasStartedStorymap`, `isIntroductionActive()`, and `startStorymapDisplay()` so the intro lifecycle is explicit instead of being inferred only from `inSlides`.
- Makes intro scroll locking apply only to single storymap pages while the intro is actually active, and pauses Scrollama step handling until the intro has started.
- Keeps navigation entry/exit state-driven through React (`isNavigating`) and makes exiting navigation return to active slide mode with `mapBrightness: 1`, `inSlides: true`, and `hasStartedStorymap: true`.
- Moves the state-driven section classes and navigation viewport CSS here (`story-map--navigating` / `story-map--intro-active`) so the post/embed branch does not need to carry competing navigation layout changes.
- Keeps the Mapbox/MapLibre navigation map controls outside the map-owned DOM and keeps the return-to-slides control positioned by the storymap container.

Why this is here:

#567 needs these same lifecycle hooks, but it should not be responsible for navigation hardening. By moving the shared navigation/intro behavior into this PR, #567 can focus on post/embed render data and REST endpoint selection.

Validation:

- Merged #565 into a fresh `upstream/develop` integration branch, then merged #567: no conflicts.
- Replayed the original local sequence `#563 -> #564 -> #565 -> #566 -> #567`: no conflicts.
- Ran `npm ci`, `npm run build`, and `npm run test:unit -- --runInBand` in the integration worktree. Build and tests passed. The build still prints existing WP-CLI/PHP deprecation warnings, but exits successfully.

leopiccionia:
The existing behavior is that, on scrolling, the storymap will advance to next slide, even if introduction is shown.

I'm understanding that the change is intentional, right?

leopiccionia:
After reading #567, I understand the navigation changes are intentional.

PR #567, opened 2026-05-20
Fix storymap display state in post contexts

bwstefano:
## Summary

This PR fixes several related storymap rendering/state issues that appear when storymaps are used inside regular posts, especially when a post contains an existing embedded storymap or multiple storymaps created directly in the post editor.

The root issue was that the storymap display component reused page-level assumptions from standalone storymap pages in embedded/post contexts. That caused global scroll locking, introduction state, navigation mode, and map brightness to leak outside the intended storymap instance.

## What changed

- Embedded storymaps now preserve the referenced storymap post ID and REST base when rendering their saved block data.
- Storymap header data is fetched from the correct REST endpoint for both `storymap` posts and regular `posts`.
- Missing or invalid embedded storymap payloads are handled defensively instead of rendering broken state.
- Introduction scroll locking is now limited to dedicated `single-storymap` pages. Embedded storymaps inside posts no longer lock the whole page.
- The introduction is now an explicit one-way state: before `START`, slides are kept collapsed and hidden; after `START`, the intro/header/button do not render again when scrolling back up.
- Scrollama handlers ignore slide transitions while the introduction is still closed, avoiding accidental map/layer changes while the reader is simply scrolling past an embedded storymap.
- Normal storymaps without introduction no longer dim the map when the reader scrolls to the page bottom and returns to the top.
- Navigation mode is scoped to the storymap area instead of behaving like a full-page overlay in posts.
- The "Back to top" control now belongs to the storymap navigation context and exits navigation with full map brightness and slide state restored.
- Storymap containers now reserve at least `100vh`, preventing following post blocks from overlapping short storymaps whose sticky map is taller than their slide content.
- Layer matching in rendered storymap data now normalizes IDs and avoids accessing missing selected-layer data.

## Design decisions

- The intro remains blocking only on standalone storymap pages, where the full-page storymap experience is expected.
- In regular posts, the post remains scrollable. If a storymap has an intro, readers can pass the intro without seeing all slides unless they click `START`.
- After `START`, the intro is considered consumed for that component instance. Scrolling back to the top should not resurrect the intro or reapply the dimmed map mask.
- The navigation map is rendered in-flow with the storymap instead of `fixed`, so it cannot cover the whole post or unrelated content.
- The `100vh` minimum is applied to the storymap shell, not to individual slides, so longer storymaps still size naturally while short storymaps reserve enough room for the sticky map.

## Local verification

Manual browser checks were run against `infoamazonia.local` before porting the fix:

- `http://infoamazonia.local/?p=229218`
  - Reproduced the intro-enabled storymap inside a post.
  - Verified the page remains scrollable before clicking `START`.
  - Verified slides are hidden/collapsed until `START`.
  - Verified `START` removes the intro permanently for that instance.
  - Verified scrolling down and back up does not bring the intro/button back and leaves the map at `brightness(1)`.

- `http://infoamazonia.local/?p=229212`
  - Reproduced the `Navigate the map` flow inside a post.
  - Verified navigation mode stays within the storymap area rather than covering the full post.
  - Verified `Back to top` exits navigation without restoring the dimmed intro mask.

- `http://infoamazonia.local/?p=229231`
  - Reproduced two storymaps created directly inside the post.
  - Verified following blocks no longer overlap short storymaps.
  - Verified scrolling to the bottom and back up leaves both maps at `brightness(1)`.

- `http://infoamazonia.local/storymap/ate-a-ultima-gota-colombia/`
  - Verified standalone storymaps still lock page scroll on the initial introduction.
  - Verified clicking `START` restores normal slide display and prevents the intro from returning.

## Automated checks

- `php -l src/includes/class-jeo.php`
- `npm run build`
- `npm run check:wporg-compliance`
- `git diff --check`

`npm run build` completed successfully. WP-CLI emitted existing PHP deprecation warnings while generating i18n JSON files, but the build finished successfully.

### DIFF
--- a/src/includes/class-jeo.php
+++ b/src/includes/class-jeo.php
@@ -718,13 +718,49 @@ class Jeo {
 	 */
 	public function embedded_story_map_dynamic_render_callback( $block_attributes, $content ) {
 		$content = json_decode( $this->extract_json_from_block_content( $content ) );
+		if ( ! is_object( $content ) || empty( $content->attributes->storyID ) ) {
+			return '';
+		}
 
-		$story_id                       = $content->attributes->storyID;
-		$story                          = get_post( $story_id );
-		$story_block                    = parse_blocks( $story->post_content )[0];
+		$story_id = absint( $content->attributes->storyID );
+		$story    = get_post( $story_id );
+		if ( ! $story instanceof WP_Post ) {
+			return '';
+		}
+
+		$story_blocks = parse_blocks( $story->post_content );
+		$story_block  = $story_blocks[0] ?? null;
+		if ( ! is_array( $story_block ) || empty( $story_block['attrs'] ) ) {
+			return '';
+		}
+
 		$story_block['attrs']['postID'] = $story_id;
 
 		return $this->story_map_dynamic_render_callback( $block_attributes, wp_json_encode( $story_block['attrs'] ) );
+	}
+
+	/**
+	 * Add REST endpoint information for the post shown in the story map header.
+	 *
+	 * @param object $saved_data Story map saved data.
+	 * @param int    $post_id    Post ID to expose.
+	 * @return void
+	 */
+	private function set_story_map_post_rest_data( $saved_data, int $post_id ) {
+		if ( ! $post_id ) {
+			return;
+		}
+
+		$post_type = get_post_type( $post_id );
+		if ( ! $post_type ) {
+			return;
+		}
+
+		$post_type_object = get_post_type_object( $post_type );
+		$rest_base        = ! empty( $post_type_object->rest_base ) ? $post_type_object->rest_base : $post_type;
+
+		$saved_data->postID       = $post_id;
+		$saved_data->postRestBase = $rest_base;
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


bwstefano:
Follow-up note on commit `afe08788`:

This extra commit fixes the PHPCS/WPCS failure reported by CI after the initial PR push.

The storymap render data intentionally needs to expose camelCase keys (`postID` and `postRestBase`) because those names are part of the JSON contract consumed by the storymap JavaScript component. WPCS flags direct PHP object-property access with camelCase names, so the commit switches those reads/writes to dynamic property access, for example:

```php
$saved_data->{'postID'}
```

This preserves the serialized JSON shape expected by the frontend.

It also adds the missing docblock entry for the optional `$block` argument accepted by WordPress render callbacks.

There is no intended behavior change in this follow-up commit; it only makes the PHP implementation comply with the project coding standards while keeping the frontend contract intact.

Validated locally with:

- `vendor/bin/phpcs --standard=phpcs.xml.dist`
- `php -l src/includes/class-jeo.php`
- `git diff --check`

bwstefano:
I pushed `6029f19a` (`Narrow storymap post embed display changes`) to keep this PR focused on post/embed behavior and remove the navigation-layout overlap with #565.

What this commit does:

- Drops the competing storymap navigation SCSS from this branch so #565 owns the `story-map--navigating` / navigation viewport layout work.
- Reuses the shared intro/navigation helpers from #565 (`isIntroductionActive()`, `startStorymapDisplay()`, and state-driven navigation entry/exit) instead of carrying a separate version here.
- Keeps this PR focused on the post/embed fix: `postRestBase`, configurable REST fetching, guarded `response.ok` handling, and safe rendering when post title/date data is not available yet.
- Keeps the PHP render-data changes intact so embedded story maps expose the correct `postID` and REST base for the post being displayed.
- Normalizes a few tiny formatting/handler differences, including removing an unawaited `sleep(1000)` wrapper around the return-to-slides click handler.

Why this is here:

The previous version of this PR touched the same navigation regions as #565, which caused the integration conflict in `storymap-display.js` and `storymap-display.scss`. After this commit, #567 should layer cleanly on top of #565 and only add the post/embed REST context behavior.

Validation:

- Merged #565 into a fresh `upstream/develop` integration branch, then merged #567: no conflicts.
- Replayed the original local sequence `#563 -> #564 -> #565 -> #566 -> #567`: no conflicts.
- Ran `npm ci`, `npm run build`, and `npm run test:unit -- --runInBand` in the integration worktree. Build and tests passed. The build still prints existing WP-CLI/PHP deprecation warnings, but exits successfully.

PR #570, opened 2026-05-20
Stabilize storymap rich text editing panels

bwstefano:
## Summary

- Keep the story settings description editor and slide content editor at stable heights, with long text scrolling inside each CKEditor instead of growing the surrounding panel.
- Increase the slide title editor height so longer titles are not cramped or internally clipped.
- Preserve the Gutenberg editor scroll position while typing, pressing Enter, or deleting inside storymap CKEditor instances so the map canvas and panels do not jump while editing.
- Version the storymap editor stylesheet from the generated `mapBlocks.asset.php` metadata so local/editor CSS updates are cache-busted together with the block script.

## Why

Editing multi-line storymap text was uncomfortable because CKEditor content changes could cause the Gutenberg editor scroller to re-anchor around the active caret. In practice, pressing Enter in slide content or story settings moved the storymap canvas and controls instead of keeping the panel stationary while the text changed.

The title field also used a very small fixed height, which made the first storymap slide title field feel undersized. The CSS-only local test confirmed that fixed-height CKEditor areas with internal scrolling remove the growth/shrink behavior, and the remaining movement came from Gutenberg scroll adjustments. This patch contains both parts of the fix.

## What changed

- Added a shared `stabilizeStorymapEditorScroll()` setup for CKEditor instances that captures the Gutenberg content scroller before text input and restores it immediately after CKEditor updates while focus remains in the same editor.
- Applied that scroll stabilization to the story settings description editor and the slide content editor.
- Added wrapper classes for the story description and slide content editors so their layout rules are explicit.
- Set the story description and slide content editors to fixed 300px editing areas with internal scrollbars.
- Raised the slide title editor to 72px and removed its previous 44px max-height limit.
- Made the slides settings container taller and viewport-aware so the expanded editing controls have enough room.
- Registered the map blocks editor stylesheet with the generated asset version, matching the script cache-busting behavior.

## Validation

- Tested manually in the local `infoamazonia.local` storymap editor, including multi-line typing and deletion in slide content and story settings fields.
- Verified the Gutenberg editor scroll position stayed stable while CKEditor internal scroll handled long content.
- Ran `npm run build` successfully. The command emits existing WP-CLI PHP deprecation warnings in the local environment, but webpack and i18n generation completed successfully.
- Ran `npm run test:unit -- --runInBand` successfully: 14 suites and 43 tests passed.

### DIFF
--- a/src/includes/class-jeo.php
+++ b/src/includes/class-jeo.php
@@ -593,7 +593,7 @@ class Jeo {
 			'jeo-map-blocks',
 			JEO_BASEURL . '/js/build/mapBlocks.css',
 			array( 'mapgl', 'mapgl-react-style' ),
-			JEO_VERSION,
+			$map_blocks_assets['version'] ?? JEO_VERSION,
 		);
 		wp_register_script(
 			'jeo-map-blocks',
--- a/src/js/src/map-blocks/storymap-editor.js
+++ b/src/js/src/map-blocks/storymap-editor.js
@@ -1121,7 +1121,10 @@ export default function StoryMapEditor ( { attributes, setAttributes } ) {
 																	editor={ ClassicEditor }
 																	data={ slide.title }
 																	config={ titleEditorConfig }
-																	onReady={ ( editor ) => setupEditor( editor, { singleLine: true } ) }
+																	onReady={ ( editor ) => setupEditor( editor, {
+																		singleLine: true,
+																		stabilizeOuterScroll: true,
+																	} ) }
 																	onChange={ ( event, editor ) => {
 																		setCurrentSlideIndex( slideIndex );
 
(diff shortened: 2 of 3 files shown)


bwstefano:
Added a follow-up commit for the slide title editor behavior: `7aba66fb` (`Let storymap slide titles grow while editing`).

This keeps the title editor aligned with the scroll-stability work already done for the slide content editor. The title field now:

- expands naturally from its existing 72px minimum height instead of clipping longer titles;
- caps at 144px so very long titles do not make the slide panel push the editor viewport around;
- uses internal vertical scrolling after that cap;
- opts into the same outer-scroll stabilization used by the content editor, so typing in the title should not cause the WordPress editor canvas to jump up or down.

PR #572, opened 2026-05-20
Optimize storymap admin listing memory usage

bwstefano:
## Summary

This PR optimizes the Story Map admin listing so large production datasets can open `/wp-admin/edit.php?post_type=storymap` without exhausting PHP memory.

The fix keeps the WordPress Screen Options value for items per page working normally, while preventing the list table from loading and parsing full Story Map block payloads for every row.

## Why this is needed

On the local InfoAmazonia copy, Story Maps are the real memory pressure point:

- the Portuguese Story Map listing has 91 visible items;
- Story Map `post_content` can be several MB per post;
- loading 50 full Story Maps at once can require well over 100 MB just for raw `post_content`, before WordPress, WPML, Yoast, Co-Authors Plus, and other admin integrations add overhead.

The visible fatal error was:

```text
Allowed memory size of 268435456 bytes exhausted ... wp-includes/class-wpdb.php
```

That `wpdb` failure was a symptom: WordPress was materializing very large post rows into memory for an admin screen that only needs lightweight listing data.

## Root cause

There were several overlapping contributors:

1. Story Maps are registered as hierarchical posts, so the admin list can trigger page-like behavior that requests all IDs/parents and bypasses normal pagination.
2. The main admin list query selected complete post rows, including huge `post_content` values.
3. Core list-table code such as `wp_check_post_lock()` can call `get_post( $id )` while rendering each row. If the lightweight row is not cached, WordPress reloads the full post row by ID.
4. Yoast SEO list columns recalculate meta context and parse block content from each Story Map row, which defeats any lightweight query and can trigger `parse_blocks()` on multi-MB payloads.
5. The hidden Quick Edit parent dropdown is unnecessary for JEO Story Maps and can also force expensive hierarchical post behavior.

## What changed

- Keeps the Story Map admin list paginated using the user's configured Screen Options value instead of capping it artificially.
- Marks only the Story Map admin list main query for lightweight handling.
- Replaces the main list query fields with an explicit lightweight column list and returns empty placeholders for `post_content` and `post_content_filtered`.
- Disables post meta cache priming for this listing query.
- Temporarily primes the object cache with lightweight `WP_Post` objects during the request so row-level admin callbacks do not refetch full Story Map rows.
- Clears those request-local lightweight cache entries on `shutdown` to avoid leaving partial post objects around.
- Removes `wpseo-*` columns from the Story Map list table because they force full block-content parsing in the listing.
- Disables the hidden Quick Edit parent dropdown for Story Maps, since JEO does not use parent/child Story Map relationships in the evaluated InfoAmazonia data.

## Impact

The Story Map listing should now behave like a normal paginated admin list even when users choose higher Screen Options values such as 50 or 100 items per page.

The main visible tradeoff is that Yoast SEO columns no longer appear on the Story Map list table. The Yoast metabox and SEO data on individual Story Map edit screens are not changed.

Maps and Layers are intentionally not changed here. In the evaluated InfoAmazonia copy, their `post_content` payloads were tiny compared with Story Maps, so applying the same content-light query to those post types would add unnecessary behavioral risk.

## Validation

Validated against the local InfoAmazonia copy at `infoamazonia.local`.

Story Map admin listing:

| Screen Options value | Result |
| --- | --- |
| 20 items per page | `HTTP 200`, 20 rows, 91 items, 5 pages, no fatal error |
| 50 items per page | `HTTP 200`, 50 rows, 91 items, 2 pages, no fatal error |
| 100 items per page | `HTTP 200`, 91 rows, 91 items, 1 page, no fatal error |

Additional HTTP checks:

- Story Map search returned `HTTP 200` with no memory fatal.
- Draft status filter returned `HTTP 200` with no memory fatal.
- WPML language filter for `lang=en` returned `HTTP 200` with no memory fatal.
- Editing an individual Story Map returned `HTTP 200` and still loaded the full edit screen.
- A published Story Map frontend URL returned `HTTP 200`.
- Quick Edit buttons still render.
- The parent dropdown is absent from the Story Map Quick Edit markup.
- Yoast table columns are absent from the Story Map list table.

Local code checks:

```text
php -l src/includes/storymap/class-storymap.php
vendor/bin/phpcs --standard=phpcs.xml.dist src/includes/storymap/class-storymap.php
php scripts/check-php-compat.php
git diff --check -- src/includes/storymap/class-storymap.php
```

### DIFF
--- a/src/includes/storymap/class-storymap.php
+++ b/src/includes/storymap/class-storymap.php
@@ -27,6 +27,13 @@ class Storymap {
 	public $post_type = 'storymap';
 
 	/**
+	 * Post IDs cached with lightweight admin-list objects during this request.
+	 *
+	 * @var int[]
+	 */
+	private $lightweight_admin_list_cache_ids = array();
+
+	/**
 	 * Register storymap hooks.
 	 *
 	 * @return void
@@ -37,6 +44,13 @@ class Storymap {
 		add_action( 'admin_init', array( $this, 'add_capabilities' ) );
 
 		add_action( 'pre_get_posts', array( $this, 'show_on_archives' ) );
+		add_action( 'pre_get_posts', array( $this, 'prepare_admin_list_query' ) );
+		add_filter( 'request', array( $this, 'paginate_admin_list_request' ) );
+		add_filter( 'posts_fields', array( $this, 'lightweight_admin_list_fields' ), 10, 2 );
+		add_filter( 'the_posts', array( $this, 'prime_lightweight_admin_list_cache' ), 10, 2 );
+		add_filter( 'manage_' . $this->post_type . '_posts_columns', array( $this, 'remove_expensive_admin_list_columns' ), 20 );
+		add_filter( 'quick_edit_dropdown_pages_args', array( $this, 'disable_admin_parent_dropdown' ) );
+		add_action( 'shutdown', array( $this, 'clear_lightweight_admin_list_cache' ) );
 
 		add_filter( 'rest_prepare_storymap', array( $this, 'prepare_rest_response' ), 10, 2 );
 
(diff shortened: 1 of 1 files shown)


leopiccionia:
Why are storymaps hierarchical, though? Is it useful for any project?

leopiccionia:
Note: Check if this affects WPML before merging.

**Edit:** It works with WPML.

PR #574, opened 2026-05-21
Standardize the i18n build and release flow

### LINKED ISSUE #522: Call wp i18n make-json directly from package.json
This follow-up comes from the review discussion on PR #515 ("Improve JEO i18n copy and release flow").

Related commits:
- 7a617ddab5e28e890cb37c3735b8ebb7bc4efa8e (`Improve JEO i18n copy and release flow`)
- a66eaae3e66a2f85cc5143b7909d9ffe4b4d9c82 (`Localize editor UI and polish JEO labels`)
- 70d14995b47eabfe4e49b00bd0a71fc0427a8523 (`Install WP-CLI in Node frontend workflow`)

PR #515 added `scripts/make-i18n-json.mjs`, and `package.json` now invokes that Node wrapper to run `wp i18n make-json`.

@leopiccionia questioned why this is the only WP-CLI i18n command being invoked through a dedicated Node script instead of being called directly from `package.json`.

That concern makes sense: right now the wrapper mostly exists to:
- read `WP_CLI_BIN`
- print a friendlier error if `wp` is missing from `PATH`
- set a Windows-specific `shell` option

Those benefits are relatively small compared to the extra indirection, especially if we continue moving more i18n-related automation into `package.json`.

## Proposal

Replace the wrapper with a direct command in `package.json`, for example:

```json
"i18n:json": "wp i18n make-json src/languages --no-purge"
```

If we still want custom messaging or cross-platform handling, we should either:
- document that explicitly and keep the wrapper on purpose, or
- standardize the same wrapper approach for all WP-CLI i18n commands instead of only this one

## Why track this separately

This is mostly a maintainability/consistency cleanup. The current implementation works, but the extra Node layer is surprising and makes the i18n build flow a bit harder to scan.

bwstefano:
## Summary

- Replaces the custom `scripts/make-i18n-json.mjs` wrapper with direct WP-CLI i18n scripts in `package.json`.
- Adds explicit scripts for refreshing `jeo.pot`, updating PO catalogs, compiling MO catalogs, and generating JS JSON catalogs.
- Moves `src/languages/*.mo` to generated release artifacts, matching the existing generated-only policy for `src/languages/*.json`.
- Updates the README with a two-phase translation workflow: refresh source catalogs, review PO translations, then compile runtime artifacts.

## Why

Closes #522.

The current workflow only documented JavaScript JSON generation as a build artifact. MO catalogs were still versioned, which made PO updates easy to desynchronize from the WordPress runtime files and left maintainers guessing which generated files should be committed.

This PR makes the policy explicit:

- `src/languages/jeo.pot` and `src/languages/*.po` remain versioned source files.
- `src/languages/*.mo` and `src/languages/*.json` are generated during build/release.
- Human review remains between source catalog updates and compiled runtime artifacts.

`npm run build` and `npm run build:release` now compile both MO and JSON catalogs, so release packages still include runtime translations even though those generated files are no longer committed to Git.

## Validation

- `npm ci`
- `npm run i18n:compile`
- `npm run build`
- `npm run build:report`
- `msgfmt --check --statistics -o /tmp/jeo-es_CO.mo src/languages/jeo-es_CO.po`
- `msgfmt --check --statistics -o /tmp/jeo-es_ES.mo src/languages/jeo-es_ES.po`
- `msgfmt --check --statistics -o /tmp/jeo-pt_BR.mo src/languages/jeo-pt_BR.po`

WP-CLI still emits the known local PHP deprecation noise, but the i18n commands complete successfully and generate the expected MO/JSON artifacts.


package.json
```diff
@@ -17,8 +17,13 @@
 		"check:wporg-compliance": "node scripts/check-wporg-compliance.mjs",
 		"start": "wp-scripts start",
 		"build:assets": "wp-scripts build && node scripts/patch-build-compliance.mjs",
-		"i18n:json": "node scripts/make-i18n-json.mjs",
-		"build": "npm run build:assets && npm run i18n:json",
+		"i18n:pot": "wp i18n make-pot src src/languages/jeo.pot --domain=jeo --exclude=assets",
```
leopiccionia:
```suggestion
		"i18n:pot": "wp i18n make-pot src src/languages/jeo.pot --domain=jeo --exclude=js/src",
```

leopiccionia: O JEO não tem uma pasta `assets`. No entanto, o conteúdo da pasta `js/src` pode ser excluído, pois apenas o conteúdo da pasta `js/build` é incluído.

leopiccionia (changes requested):

package.json
bwstefano: Faz sentido. Vou aplicar sua sugestão.

PR #575, opened 2026-05-21
Add CI validation for WordPress language files

### LINKED ISSUE #521: Add CI validation for WordPress language files with Test WordPress language files
## Summary

This issue proposes adopting the [`Test WordPress language files`](https://github.com/marketplace/actions/test-wordpress-language-files) GitHub Action as a dedicated CI check for the JEO plugin.

The goal is to automatically verify that our WordPress internationalization files stay up to date whenever translatable strings change.

## Why

JEO already ships and tracks WordPress language assets, including:

- a versioned `src/languages/jeo.pot`
- versioned PO/MO files such as `jeo-pt_BR.po` and `jeo-es_CO.po`
- generated JavaScript translation JSON files under `src/languages/`

We already generate JavaScript translation JSON catalogs during the build, but we do not currently have a dedicated CI check that verifies whether the POT and PO files are still in sync with the source code after strings change.

That gap creates a few risks:

- new or updated strings can land without the POT file being refreshed
- PO files can drift behind the source over time
- translation maintenance becomes more reactive and release-driven than review-driven
- contributors do not get fast feedback when i18n source files need regeneration

## What this action gives us

The `Test WordPress language files` action runs the official WordPress CLI i18n commands to:

- regenerate the POT file
- update PO files
- fail when those files are out of date
- fail on warnings by default
- produce a patch artifact and warnings output for easier review

That means we would gain a lightweight, WordPress-native guard rail that fits well with the plugin’s existing localization workflow.

## Why this is a good fit for JEO

This action complements the current repository setup rather than replacing it.

It does not replace:

- asset builds
- JavaScript translation JSON generation
- release packaging
- Plugin Check
- smoke tests

Instead, it fills a missing validation layer: ensuring that our tracked translation source files stay current as part of normal CI.

For JEO specifically, the benefits would be:

- earlier detection of stale translation sources in pull requests
- less manual POT/PO maintenance before releases
- better confidence that translators are working from current extraction data
- fewer “translation drift” surprises when strings change in PHP, JS, block metadata, or other translatable sources
- clearer contributor feedback when a patch needs i18n file updates

## Proposal

Add a dedicated workflow, or a dedicated job in an existing validation workflow, that runs the `Test WordPress language files` action against the plugin sources.

Suggested initial behavior:

- run on `pull_request`
- run on `push`
- fail the workflow when POT or PO files are outdated
- fail the workflow on warnings
- target the plugin source tree under `src/`
- target the existing `src/languages/` directory

## Suggested implementation details

The workflow should be configured to match the repository structure:

- `slug`: `jeo`
- `source`: `src`
- `languages_directory`: `languages`
- `pot_file`: `src/languages/jeo.pot`
- `po_path`: the existing tracked PO files under `src/languages/`

We should also review whether the default ignored header patterns are sufficient for our tracked POT file headers, especially around generated timestamps.

## Expected outcome

After this is in place:

- a PR that changes translatable strings without updating POT/PO files will fail CI
- maintainers will get a machine-readable patch showing the required i18n updates
- localization upkeep will move earlier in the development lifecycle
- release preparation should become cleaner because the repository will be less likely to accumulate stale i18n metadata

## Non-goals

This proposal is not intended to:

- replace our current build pipeline
- replace JSON catalog generation
- replace release-time packaging or deployment
- replace broader localization process improvements

## Acceptance criteria

- a GitHub Actions workflow uses `Test WordPress language files`
- the workflow runs on pull requests and pushes
- it validates the JEO plugin sources under `src/`
- it checks the tracked POT and PO files in `src/languages/`
- CI fails when the language files are outdated
- CI fails when the action reports warnings, unless we explicitly decide otherwise
- the workflow output is understandable enough for contributors to fix the issue quickly

## Notes

This would be a targeted, low-risk improvement to CI quality. It is especially valuable in a repository like JEO where translation source files are tracked and should stay aligned with ongoing PHP and JavaScript changes.

bwstefano:
## Summary

- Add a dedicated WordPress Language Files workflow using `holyhope/test-wordpress-languages-github-action@v4.0.2`.
- Run the check on pull requests, pushes, and manual dispatches under PHP 8.4.
- Validate `src/languages/jeo.pot` and the tracked `src/languages/jeo-*.po` files against the plugin sources under `src/`.
- Fail on language-file diffs and extraction warnings, and upload the generated patch/warnings artifacts to make failures easier to review.

Closes #521.

## Implementation Notes

The workflow leaves `po_path` unset intentionally. In v4.0.2 the action auto-discovers PO files adjacent to the configured POT file, which keeps the workflow independent from the active locale list and avoids hard-coding `es_CO`, `es_ES`, or `pt_BR` into CI.

This PR only adds the CI guardrail. It intentionally does not refresh `jeo.pot` or PO files against the current `develop` snapshot, because the intended release order has #568 and #574 landing first. Those PRs change the translation baseline by removing the deprecated `es_ES` source catalog and standardizing generated i18n release artifacts.

After #568 and #574 are merged, this branch should be updated from `develop` and the workflow should be rerun against that new base. If the action still reports POT/PO drift at that point, the required catalog refresh should be added as a follow-up commit calculated from the final i18n baseline.

## Validation

- `ruby -ryaml -e 'YAML.load_file(".github/workflows/wordpress-languages.yml"); puts "YAML ok"'`
- The GitHub Actions workflow itself previously passed when run with refreshed catalogs; after dropping the temporary catalog refresh commit, the new check may fail against the current `develop` until #568 and #574 establish the intended translation baseline.


leopiccionia (changes requested):
It's failing because POT timestamp headers are different. Copilot's suggestions didn't fix the issue.

PR #584, opened 2026-05-22
Update dependencies and Jest ESM handling

bwstefano:
## Summary

This PR refreshes the project dependency set and the Jest configuration after the dependency maintenance pass we tested locally.

It updates the JavaScript package ranges for the editor/runtime stack, including CKEditor 5, the React CKEditor integration, MapLibre-related packages, RJSF, Eta, `react-map-gl`, and the WordPress editor/plugin packages. It also upgrades `@wordpress/scripts` to the current 32.x line and regenerates `package-lock.json` from that dependency graph.

It removes direct npm dependencies that are no longer needed as project-level dependencies after the tooling update: `@wordpress/edit-post`, `ajv`, `css-loader`, `postcss`, `sass`, `sass-loader`, `style-loader`, and `webpack`. The build still gets the tooling it needs from `@wordpress/scripts`; keeping those packages as direct dependencies was making the project look responsible for packages it does not import or configure directly.

It also adds explicit direct dependencies for the WordPress packages and `date-fns` modules that the source imports by package name. During the cleanup, relying on incidental transitive hoisting became visibly fragile: tests could no longer resolve `@wordpress/components` or `date-fns/locale` in a clean install even though those packages were still present deeper in the WordPress package tree. Declaring the packages we import directly makes the dependency contract clear and keeps npm layout changes from breaking resolution.

The remaining overrides were simplified:

- `react-autosize-textarea` stays pinned to the local metadata-only override because Gutenberg still tracks the abandoned package upstream and the replacement path remains unresolved.
- `webpack-dev-server` stays overridden to `^5.2.4`, matching the tested security/maintenance update while Gutenberg's own Dependabot PR is still open.
- `markdownlint-cli` now overrides the package itself to `^0.48.0` instead of only forcing its nested `minimatch`.
- `copy-webpack-plugin` now overrides the package itself to `^14.0.0` instead of only forcing its nested `serialize-javascript`.
- Older targeted overrides for `js-yaml` and `@typescript-eslint/typescript-estree` were removed because the refreshed dependency tree no longer needs them.

On the PHP tooling side, this updates `dealerdirect/phpcodesniffer-composer-installer` from `v1.2.0` to `v1.2.1` in `composer.lock` without changing the project Composer requirements.

Finally, Jest now transforms nested ESM `uuid` packages as well as the existing ESM packages already listed in the config. The newer WordPress dependency graph can place ESM `uuid` under nested package paths, and the previous transform exception only handled the top-level layout.

## Motivation and decisions

The goal was to reduce custom maintenance and stale direct dependencies while staying aligned with the WordPress package ecosystem. The cleanup intentionally avoids broad runtime migrations such as React 19 or `react-leaflet` 5, because the current WordPress packages are still on React 18 and those upgrades would require a separate compatibility pass.

I kept overrides only where they address a concrete, current gap that upstream has not resolved yet. Where a package-level override was available and tested (`markdownlint-cli`, `copy-webpack-plugin`), I replaced nested dependency surgery with the newer package version instead.

I also chose to make the WordPress package imports explicit rather than relying on `@wordpress/editor` or another package to hoist them. That makes the lockfile less sensitive to npm layout changes and more accurately documents what the plugin imports.

## Validation

- `npm ci`
- `npm run build`
- `npm run build:report`
- `npm audit --audit-level=high`
- `composer audit --locked`

`npm audit --audit-level=high` passes. The remaining audit output is limited to the known moderate issues in `showdown` through WordPress packages and `uuid` through `sockjs`/`webpack-dev-server`; neither currently has a non-breaking fix in this dependency graph. The local build still prints WP-CLI PHP deprecation warnings from the installed WP-CLI PHAR, but the build completes successfully.


package.json
```diff
@@ -42,49 +42,48 @@
...
+		"@wordpress/data": "^10.46.0",
+		"@wordpress/dom-ready": "^4.46.0",
+		"@wordpress/editor": "^14.46.0",
+		"@wordpress/element": "^6.46.0",
+		"@wordpress/html-entities": "^4.46.0",
+		"@wordpress/i18n": "^6.19.0",
+		"@wordpress/icons": "^13.1.0",
+		"@wordpress/media-utils": "^5.46.0",
+		"@wordpress/plugins": "^7.46.0",
+		"@wordpress/url": "^4.46.0",
```
leopiccionia: `@wordpress/editor` already contains all these other `@wordpress/*` packages indirectly (in the case of `@wordpress/dom-ready`, that's a dependency of `@wordpress/a11y`) or directly (in all other cases).
https://www.npmjs.com/package/@wordpress/editor?activeTab=dependencies

I think mainting `@wordpress/editor` in `dependencies` and `@wordpress/scripts` in `devDependencies` will diminish the burden of handling dependencies mismatches.

bwstefano: Makes sense. I’ll add a commit to reduce the explicit dependency surface back to `@wordpress/editor` plus `@wordpress/scripts` so the Gutenberg package versions stay aligned and easier to maintain.

PR #601, opened 2026-06-09
Fix public map bounds and storymap layer visibility

bwstefano:
## Summary

This PR fixes the public map/storymap regressions observed after installing the current `jeo.zip` on InfoAmazonia.

- Restores compatibility with legacy `pan_limits` values saved in the older coordinate order.
- Uses the initial public map center to disambiguate bounds that are technically valid in both the standard and legacy formats.
- Prevents storymap layer visibility changes from running before the Mapbox/MapLibre style and layers are ready.
- Exposes a stable storymap map container class contract for themes: `.jeo-storymap-map` / `.js-jeo-storymap-map`.
- Documents that themes should use the JEO-owned storymap container class instead of runtime classes such as `.mapboxgl-map` or `.maplibregl-map`.

### Invalid LngLat latitude errors on public maps

Some public maps had `pan_limits` metadata that was still effectively stored in the legacy order. After the recent public map initialization changes, those values were interpreted as Mapbox/MapLibre bounds directly, producing invalid latitude values such as `-108` and causing:

```text
Invalid LngLat latitude value: must be between -90 and 90
```

The fix validates both possible bounds interpretations and chooses the legacy interpretation when the standard interpretation is invalid.

### Public map center clamped away from editor-defined center

Some maps had bounds that were valid in both interpretations, but only the legacy interpretation contained the editor-defined initial center. The public map could therefore initialize with a different apparent center than the editor.

The fix passes the initial center into `getPanLimitsMaxBounds()` and prefers the bounds interpretation that contains it.

### Storymap `Style is not done loading` errors

Storymap scroll events can fire while MapLibre is still loading its style. The previous implementation called `setLayoutProperty()` directly from scroll handlers, which can throw:

```text
Style is not done loading
```

The fix centralizes storymap layer visibility updates behind guards that check style readiness and layer existence before mutating layer layout.

## Theme integration note

The separate theme-side error below is intentionally not fixed in this plugin PR:

```text
Cannot read properties of null (reading 'style')
```

That error comes from `jeo-theme` querying Mapbox runtime classes before the map runtime has initialized. This PR adds a stable plugin-owned hook that themes can target instead:

```html
<div class="story-map-element jeo-storymap-map js-jeo-storymap-map"></div>
```

The actual theme fix is tracked in InfoAmazonia/jeo-theme#512.

## Validation

- `npm run test:unit -- --runTestsByPath src/js/src/jeo-map/pan-limits.test.js`
- `npm run build`
- Confirmed on InfoAmazonia after purging cached JS assets that the plugin-side public map and storymap errors no longer reproduce. The remaining `Cannot read properties of null (reading 'style')` error is the expected theme follow-up.

Note: `npm run build` still prints WP-CLI/vendor PHP deprecation warnings from the local toolchain, but exits successfully.


src/js/src/jeo-storymap/storymap-display.js
```diff
@@ -244,6 +246,73 @@ class StoryMapDisplay extends Component {
...
+			this.map.setLayoutProperty( layerSlug, 'visibility', visibility );
+		} catch ( error ) {
+			const message = error?.message || '';
+
+			if (
+				message.includes( 'Style is not done loading' ) ||
+				message.includes( 'does not exist in the map' )
+			) {
+				return;
+			}
```
leopiccionia: This code looks fragile. Has it been tested in both MapboxGL and MapLibre GL?

leopiccionia (approved):

src/js/src/jeo-storymap/storymap-display.js
bwstefano: Good point. I asked Codex to review. This is what it did in the couple of commits below:

It extracted the layer visibility guard into a small helper and added unit coverage for the shared GL JS map API used by both Mapbox GL JS and MapLibre GL JS: `isStyleLoaded()`, `getLayer()`, and `setLayoutProperty()`. The tests cover loaded/unloaded styles, missing layers, transient style-loading errors, and unexpected errors.

I'll test a little bit more in a couple of hours.

bwstefano: Tested without errors, @leopiccionia. Can we merge this to `develop`?

PR #602, opened 2026-06-09
Improve Discovery UX, loading, and performance

bwstefano:
## Summary

This PR improves the Discovery experience by fixing story-panel behavior, making filters more usable, aligning visual states with the plugin's theme-aware styling, and reducing avoidable rendering/network work on the Discovery page.

The changes are intentionally scoped to the JEO plugin. They avoid theme-specific workarounds while adding plugin-level classes, states, and behavior that can be reused across installations.

### Stories panel

- Fixes story thumbnails so lazy-loaded images render correctly in the Discovery stories list.
- Restores story text search and keeps the active query visible in the result summary.
- Moves the “loading more stories” feedback to the end of the list during infinite loading.
- Adds clearer result feedback, including “Displayed stories”, loading, empty, and active-filter states.
- Keeps “View in map” as a focused action inside each story card instead of making the whole card handle map navigation.
- Uses the active site locale for story dates instead of the browser locale, so Portuguese, Spanish, and English dates render consistently with WPML/Polylang.
- Adds map cursor feedback for story pins and clusters, with cleanup when leaving the target or switching away from the Stories tab.

### Filters

- Updates the tag selector to support multiple selected tags.
- Improves the tag search field alignment and selected-tag display.
- Adds direct month/year selection to the date picker, so users do not need to step month-by-month through the calendar.
- Fixes dark active/hover states on date-picker controls.

### Sidebar and map layers

- Unifies story and map-layer pagination behind infinite scroll.
- Adds loading and error states for the map-layer list.
- Improves selected-layer spacing so long map/layer names do not collide with the remove button.
- Fixes dark hover/active states on layer controls, remove-layer buttons, fullscreen controls, and zoom controls.
- Keeps selected layer text visually consistent while dragging.
- Adds accessible labels for collapsing and expanding the Discovery panel.
- Adjusts the Discovery block height so the section better fits the available viewport.

### Performance

- Filters the posts REST query to geolocated stories with `_related_point`, so the first Discovery story batch is useful immediately instead of fetching many non-geolocated posts and discarding them client-side.
- Restricts the posts REST payload with `_fields` to only the data used by Discovery.
- Replaces per-story media/category REST calls with batched `include` requests and in-memory ID caches.
- Avoids repeated state updates while hovering the same story pin or cluster.
- Prevents unnecessary story card re-renders when neither the story object nor its active state changes.
- Removes full Discovery re-renders on every map `move`/`zoom`; share/embed URLs still refresh after `moveend` when the share UI is open.

### Translations

- Adds/updates Portuguese and Spanish strings for the new Discovery filter, loading, active-filter, sidebar, and date-picker UI.
- Fixes a few existing Discovery translation typos and awkward strings in Portuguese and Spanish.

## Validation

- `git diff --check upstream/develop..HEAD`
- `npm run test:unit -- --runInBand src/js/src/discovery/blocks/stories.test.js src/js/src/discovery/blocks/date-range-filter.test.js`
- `npm run build:assets`
- Chrome validation on `https://infoamazonia.local/discovery/` with the rebuilt local plugin bundle:
  - Initial Stories load now shows 30 geolocated stories with 30 images.
  - Infinite scroll loads the next story batch.
  - Search for `vale` returns geolocated story results with images.
  - Story pin hover and cluster hover highlight the correct stories and reset cleanly.
  - Share links update after map camera changes.
  - Console remained free of warnings/errors during validation.


src/js/src/discovery/blocks/stories-helpers.js
```diff
@@ -25,3 +27,31 @@ export function resolveStoriesPage( params = {}, pageInfo = {} ) {
...
+		? window.languageParams?.currentLang
+		: '',
+	documentLocale = typeof document !== 'undefined'
+		? document.documentElement?.lang
+		: '',
+	browserLocale = typeof navigator !== 'undefined' ? navigator.language : '',
+} = {} ) {
+	return normalizeLocaleCode(
+		siteLocale || documentLocale || browserLocale || 'en-US'
+	);
```
leopiccionia, line 56: At the worst case, if locale is not loadable locally, it would default to browser's locale (or English, if browser's locale is itself not found), anyway.

src/js/src/discovery/blocks/stories.js
```diff
@@ -974,47 +1122,116 @@ class Stories extends Component {
...
+			queryParams.before ||
+			selectedTagIds.length > 0
+		);
+		const activeFilters = [];
+		const hoveredClusterPostsSet = new Set( this.state.hoveredClusterPostsId );
+		const storiesCountLabel = sprintf(
+			'%s: %d',
+			__( 'Displayed stories', 'jeo' ),
+			this.props.stories.length
+		);
```
leopiccionia, line 1187: For this and following ones, I'd rather do ideally something like this:
```js
// translators: %s is stories count
sprintf( __( 'Displayed stories: %s', 'jeo' ), this.props.stories.length );
```
or, at least, something like this:
```js
// translators: property key and value
sprintf( __( '%s: %s', 'jeo' ), __( 'Displayed stories', 'jeo' ), this.props.stories.length );
```

JEO has been used even in right-to-left languages, like Arabic, so I'd avoid assumptions about inline start and end.

src/js/src/discovery/blocks/stories.js
```diff
@@ -1160,16 +1402,27 @@ class Storie extends Component {
...
+						rel="noreferrer"
+					>
 						{ decodeEntities(story.title.rendered) }
-					</div>
+					</a>
 
 					<div className="date">{ storyDate }</div>
 					<div>
-						<small className="view-in-map">{ __( 'View in map', 'jeo' ) }</small>
+						<button
```
leopiccionia, line 1427: 👍 for using `<button>`.

leopiccionia (commented):
Is the loading spinner position intended? I'd center it to the whole sidebar.

src/js/src/discovery/blocks/stories-helpers.js
bwstefano: The locale fallback helper is meant to prioritize the active site language (set through WPML/Polylang) over the visitor browser locale, which avoids story dates staying in the browser language on translated pages. If that locale is unsupported, Intl can still fall back as you noted.

bwstefano:
> Is the loading spinner position intended? I'd center it to the whole sidebar.

There are two different states:

**1. Initial stories loading:**
image

**Infinite scrolling with incremental loading**
image

The initial stories loading indicator could be centered, while the incremental “loading more stories” indicator should remain at the end of the list so infinite scroll keeps clear visual continuity.

Perhaps we could simply remove the "Loading stories" string and horizontally center the spinner, while keeping their vertical positions for these two states? What do you think, @leopiccionia?

src/js/src/discovery/blocks/stories.js
bwstefano: Commit [1cf1336](https://github.com/InfoAmazonia/jeo-plugin/pull/602/commits/1cf1336287e5d739c9e799b3757ddf9f4ff93091) should fix it.

bwstefano:
@leopiccionia, commit [23c11c4](https://github.com/InfoAmazonia/jeo-plugin/pull/602/commits/23c11c4341c65273928adc8f39a61fa353051b06) addresses your concern over the spinner.

PR #607, opened 2026-06-09
Standardize build asset versioning

bwstefano:
## Summary

Fixes #604 by making asset versioning explicit across the plugin:

- Require the generated `*.asset.php` metadata for every bundled `src/js/build` entrypoint that is enqueued from PHP.
- Use the generated content-hash version for both JavaScript bundles and their matching CSS sidecars.
- Remove silent `JEO_VERSION` fallbacks from build-generated assets so incomplete release builds fail visibly instead of serving mixed cache versions.
- Keep `JEO_VERSION` for local static plugin assets by adding explicit `?ver=JEO_VERSION` cache busting to the map icon URLs passed through `jeoMapVars.images`.
- Leave local layer/legend/settings helper scripts on `JEO_VERSION` and keep external Mapbox SDK assets on the vendor SDK version.
- Stop appending the plugin version to administrator-configured typography stylesheet URLs, allowing those URLs to control their own cache policy.
- Build the ignored webpack outputs before the WordPress smoke matrix activates the plugin, because bootstrap now deliberately requires generated asset metadata.
- Refresh the affected POT/PO line references after the PHP enqueue changes.

## Why

Issue #604 describes a production cache-mixing risk where rebuilt packages can produce new `@wordpress/scripts` content hashes for some files while other related assets continue using the unchanged public plugin version. That can let Cloudflare or browser caches serve an incompatible mix of webpack runtime and dependent bundles.

This PR aligns the enqueued URL versions with the generated asset metadata for build outputs, while preserving plugin-version cache busting only for static plugin files that are not part of the build pipeline.

## Validation

Local checks:

- `php -l src/includes/class-jeo.php && php -l src/includes/sidebars/class-sidebars.php && php -l src/includes/layer-types/class-layer-types.php && php -l src/includes/legend-types/class-legend-types.php && php -l src/includes/loaders.php`
- `vendor/bin/phpcs --standard=phpcs.xml.dist src/includes/class-jeo.php src/includes/sidebars/class-sidebars.php src/includes/layer-types/class-layer-types.php src/includes/legend-types/class-legend-types.php src/includes/loaders.php`
- `git diff --check`
- `rg -n 'include .*js/build/.*asset\.php|\?\? JEO_VERSION|js/build/.*JEO_VERSION' src/includes || true`
- `npm run build:assets`
- `npm run test:unit -- --runInBand`

[...]

PR #608, opened 2026-06-09
Reuse built assets in WordPress smoke matrix

bwstefano:
## Summary

Addresses the follow-up from the PR #607 review comment:
https://github.com/InfoAmazonia/jeo-plugin/pull/607#discussion_r3383718042

- Builds the ignored webpack outputs once in a dedicated `build-smoke-assets` job.
- Uploads `src/js/build` as a short-lived `wordpress-smoke-assets` artifact.
- Makes the WordPress smoke matrix depend on that job and download the generated build directory before plugin activation.
- Removes the repeated `actions/setup-node`, `npm ci`, and `npm run build:assets` steps from every matrix entry.

## Why

PR #607 made the WordPress smoke tests build assets inside each matrix job so the plugin could require generated `*.asset.php` metadata during bootstrap. That fixed the missing asset metadata failure, but repeated the same Node install and webpack build six times.

This keeps the required generated assets available for every smoke job while avoiding repeated frontend dependency installation and repeated asset builds across the matrix.

## Validation

Local checks:

- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/wordpress-smoke.yml"); puts "YAML ok"'`
- `git diff --check`
- `npm run build:assets`
- `actionlint .github/workflows/wordpress-smoke.yml`

Note: `actionlint` still reports the pre-existing ShellCheck `SC2016` warning in the inline PHP script used by the MariaDB wait step. I confirmed the same warning exists on `upstream/release/3.0.0`, so this PR does not introduce it.

GitHub checks are passing on this PR. The new `Build smoke test assets` job passed, and all WordPress smoke matrix jobs passed after downloading the generated build artifact.

### DIFF
--- a/.github/workflows/wordpress-smoke.yml
+++ b/.github/workflows/wordpress-smoke.yml
@@ -6,9 +6,36 @@ on:
   workflow_dispatch:
 
 jobs:
+  build-smoke-assets:
+    name: Build smoke test assets
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v6
+
+      - uses: actions/setup-node@v6
+        with:
+          node-version: 24
+          cache: npm
+
+      - name: Install frontend dependencies
+        run: npm ci
+
+      - name: Build plugin assets
+        run: npm run build:assets
+
+      - name: Upload smoke test assets
+        uses: actions/upload-artifact@v7
+        with:
+          name: wordpress-smoke-assets
+          path: src/js/build
+          if-no-files-found: error
+          retention-days: 1
+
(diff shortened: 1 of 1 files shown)


leopiccionia:
Other CI changes that could be explored and measured (may be follow-ups, as they are less critical):
- Re-using `npm ci && npm run build` assets with other jobs (language checking, bundle size reporting), maybe through dependency between workflows
- Rewriting plugin checking job to use WP-CLI installation via marketplace action, like language checking, instead of requiring the MySQL services just for that (does it have other use?)

PR #618, opened 2026-06-12
Restore jeowp distribution slug

bwstefano:
> Post-merge update: this PR restored the WordPress.org distribution slug to `jeowp`. After WordPress.org plugin-team feedback that the text domain should also match the plugin slug, follow-up PR #619 aligns the plugin text domain and translation catalogs with `jeowp` as well.

## Summary

This PR restores the WordPress.org distribution slug from `jeo` back to `jeowp` across the release, packaging, validation, and local smoke-test paths.

It keeps the existing runtime/API namespace intact: block names, REST routes, option keys, PHP functions, script handles, CSS classes, and the `jeo` text domain are intentionally not renamed here.

## What changed

- Updates WordPress.org release/deploy configuration to stage, package, and deploy as `jeowp`.
- Updates the shared Plugin Check action default and CI artifact names to use the `jeowp` distribution slug.
- Updates the WordPress language validation slug while keeping the existing `jeo` domain and catalog filenames.
- Updates local smoke-test expectations so WordPress sees the active plugin as `jeowp/jeo.php`.
- Updates npm package metadata, release documentation, and ignored local zip artifacts for `jeowp.zip`.
- Updates translation catalog support URLs to point at the `jeowp` WordPress.org listing.

## Why

The previous slug change moved distribution/release plumbing to `jeo`, but the WordPress.org plugin listing and public plugin URL are `jeowp`. Returning the distribution slug to `jeowp` keeps release artifacts and WordPress.org deployment aligned with that listing without renaming the plugin's internal public surface.

## Validation

- `npm run test:unit`
- `npm run build:release`
- `npm run build:report`
- `node scripts/validate-release-meta.mjs`
- `vendor/bin/phpcs --standard=phpcs.xml.dist --ignore='*/.worktrees/*'`
- `vendor/bin/phpcs --standard=phpcs-compat.xml.dist --ignore='*/.worktrees/*'`
- `python3 -m mkdocs build --clean`
- `msgfmt --check --statistics -o /tmp/jeo-pt_BR.mo src/languages/jeo-pt_BR.po`
- `msgfmt --check --statistics -o /tmp/jeo-es_CO.mo src/languages/jeo-es_CO.po`
- `WP_CLI_PHP=/opt/homebrew/opt/php@8.4/bin/php WP_DIR=/tmp/jeo-plugin-wordpress-smoke WP_VERSION=7.0 WP_SITE_URL=http://127.0.0.1 WP_DB_HOST=127.0.0.1 WP_DB_NAME=wordpress WP_DB_USER=wordpress WP_DB_PASSWORD=wordpress bash scripts/wordpress-smoke.sh`
- Local WordPress Plugin Check reproduction against a staged `jeowp` package exited successfully.

## Notes

Local WP-CLI commands on PHP 8.5 emitted upstream deprecation noise, so the i18n and WordPress checks were rerun with PHP 8.4 to match CI. The local Plugin Check analyzer still reports non-blocking observations for the intentionally preserved `jeo` text domain and the pre-existing `wp` substring in the historical `jeowp` slug.

### DIFF
--- a/scripts/wordpress-smoke-check.php
+++ b/scripts/wordpress-smoke-check.php
@@ -32,8 +32,8 @@ foreach ( $required_post_types as $required_post_type ) {
 }
 
 $active_plugins = (array) get_option( 'active_plugins', array() );
-if ( ! in_array( 'jeo/jeo.php', $active_plugins, true ) ) {
-	fwrite( STDERR, "Plugin jeo/jeo.php is not active.\n" );
+if ( ! in_array( 'jeowp/jeo.php', $active_plugins, true ) ) {
+	fwrite( STDERR, "Plugin jeowp/jeo.php is not active.\n" );
 	exit( 1 );
 }
 
--- a/.github/actions/wordpress-plugin-check/action.yml
+++ b/.github/actions/wordpress-plugin-check/action.yml
@@ -4,7 +4,7 @@ description: Install WordPress, install the Plugin Check plugin, and run checks 
 inputs:
   plugin-slug:
     description: WordPress plugin slug to install and check.
-    default: jeo
+    default: jeowp
   plugin-build-dir:
     description: Built plugin directory to copy into the WordPress installation.
     required: true
--- a/.github/workflows/deploy-wordpress-org.yml
+++ b/.github/workflows/deploy-wordpress-org.yml
@@ -26,7 +26,7 @@ jobs:
           --health-retries=10
     env:
       WP_PATH: /tmp/wordpress
-      PLUGIN_BUILD_DIR: /tmp/plugin-check-build/jeo
+      PLUGIN_BUILD_DIR: /tmp/plugin-check-build/jeowp
       PLUGIN_CHECK_RESULTS: /tmp/plugin-check-results.txt
     steps:
     - uses: actions/checkout@v6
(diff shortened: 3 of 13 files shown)


scripts/wordpress-smoke-check.php
```diff
@@ -32,8 +32,8 @@
 }
 
 $active_plugins = (array) get_option( 'active_plugins', array() );
-if ( ! in_array( 'jeo/jeo.php', $active_plugins, true ) ) {
-	fwrite( STDERR, "Plugin jeo/jeo.php is not active.\n" );
+if ( ! in_array( 'jeowp/jeo.php', $active_plugins, true ) ) {
```
leopiccionia, line 35: 1. It's conventional to name plugin's main file after the slug e.g. `jeowp/jeowp.php`, although I'm not sure if it's strictly required.

2. I've just noticed this line is buggy: it should use `JEO_BASEPATH` constant or `plugin_dir_path()` function instead of hardcoded `jeowp`. It's common to WordPress rename the plugin folder (specially to something like `jeowp-1`, `jeowp-2`, if multiple versions are installed, even if disabled). As plugins can be deleted from disk, `jeowp-1` doesn't presuppose the existance of `jeowp`.

I'm not sure if it can affect the smoke test, though.

leopiccionia (approved):

Issue #622, opened 2026-06-15
Fix store audit issues


leopiccionia:
- [x] The URL(s) declared in your plugin seems to be invalid or does not work.
- [x] Use wp_enqueue commands
- [x] Callback calls to a function/method whose return values are output must be properly escaped
- [x] High position for your admin dashboard menu item
 - [x] Nonces and User Permissions Needed for Security
 - [x] Generic function/class/define/namespace/option names

Other possible issues (detected by AI):

- [x] The filter removes the REST per_page maximum and allows -1, enabling unbounded layer collection responses that can cause avoidable performance and resource issues.

PR #623, opened 2026-06-16
Address WordPress.org review follow-up items

bwstefano:
## Summary

Refs #622.

This PR addresses the WordPress.org review follow-up items for JEO 3.0.0 review readiness.

## Scope

1. Use the canonical GNU GPLv3 license URL for the `GPL-3.0-only` plugin metadata.
2. Keep the `jeo-map` inline style handle from `d5495fb8` and validate that pages now emit `jeo-map-inline-css` instead of the previous generic custom CSS handle.
3. Harden rendered content escaping for story maps and map layer content.
4. Move the JEO admin menu out of the core admin group by relying on the map editing capability and default plugin menu positioning.
5. Restore native REST pagination limits for `wp/v2/map-layer`, while keeping the dedicated JEO route for fetching selected map layers by ID.
6. Enforce WordPress-like visibility behavior for map and story map embeds, including missing posts, non-public posts, disabled embeds, and password-protected content.
7. Prefix discovery runtime globals with `jeowp` and remove stale build source maps during release asset compliance patching so packages do not retain old global names in ignored generated artifacts.

Note: the branch is based on `develop`, where item 2 already exists as `d5495fb8`. Because of that, GitHub's commit comparison against `develop` shows the six new commits in this branch; the PR scope still covers and validates all seven items above.

## Validation

- `npm run build`
- `vendor/bin/phpcs --standard=phpcs.xml.dist '--ignore=.worktrees/*'`
- `php -l` on touched PHP files
- `git diff --check`
- Local HTTP checks against the representative `infoamazonia.local` pages from #622
- Playwright smoke checks for maps, storymaps, embeds, and discovery globals

Observed local browser noise was limited to existing theme-side JS errors, mixed-content warnings, and external Mapbox tile 404s; plugin assets loaded successfully.

### DIFF
--- a/src/includes/class-jeo.php
+++ b/src/includes/class-jeo.php
@@ -1330,6 +1330,23 @@ class Jeo {
 	}
 
 	/**
+	 * Stop or defer rendering for embeds that need a special response.
+	 *
+	 * @param WP_Post $post Embed post.
+	 * @return void
+	 */
+	private function enforce_embed_rendering_rules( WP_Post $post ): void {
+		if ( $this->is_embed_disabled( $post ) ) {
+			wp_safe_redirect( home_url() );
+			exit();
+		}
+
+		if ( post_password_required( $post ) ) {
+			$this->render_embed_password_form( $post );
+		}
+	}
+
+	/**
 	 * Serve the public embed templates for maps, story maps, and discovery.
 	 *
 	 * @return void
@@ -1361,14 +1378,7 @@ class Jeo {
 
 			setup_postdata( $post );
 
-			if ( $this->is_embed_disabled( $post ) ) {
-				wp_safe_redirect( home_url() );
-				exit();
-			}
-
-			if ( post_password_required( $post ) ) {
-				$this->render_embed_password_form( $post );
-			}
+			$this->enforce_embed_rendering_rules( $post );
 
 			add_filter( 'the_content', array( $this, 'storymap_content' ), 1 );
 
@@ -1390,14 +1400,7 @@ class Jeo {
 			$this->wp_die_embed_not_available();
 		}
 
-		if ( $this->is_embed_disabled( $map_post ) ) {
-			wp_safe_redirect( home_url() );
-			exit();
-		}
-
-		if ( post_password_required( $map_post ) ) {
-			$this->render_embed_password_form( $map_post );
-		}
+		$this->enforce_embed_rendering_rules( $map_post );
 
 		$map_meta = get_post_meta( $map_id );
 		$args     = array();
--- a/src/includes/class-jeo.php
+++ b/src/includes/class-jeo.php
@@ -1246,90 +1246,210 @@ class Jeo {
 	}
 
 	/**
+	 * Determine whether a post can be rendered through a public embed request.
+	 *
+	 * @param WP_Post $post Post requested by the embed endpoint.
+	 * @return bool
+	 */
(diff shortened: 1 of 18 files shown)


src/includes/class-jeo.php
```diff
@@ -1210,91 +1245,211 @@ public function register_embed_query_var( $vars ) {
...
 
-					$args['per_page'] = 1;
-					$request          = new WP_REST_Request( 'GET', '/wp/v2/posts' );
-					$request->set_query_params( $args );
-					$response = rest_do_request( $request );
-					$server   = rest_get_server();
-					$data     = $server->response_to_data( $response, false );
+			if ( post_password_required( $post ) ) {
+				$this->render_embed_password_form( $post );
+			}
```
leopiccionia: Can we refactor this into a function? This is almost identical to lines 1388-1400.

leopiccionia (approved):
I'm not sure if this addresses all cases of "Generic function/class/define/namespace/option names", but otherwise LGTM.

bwstefano:
> I'm not sure if this addresses all cases of "Generic function/class/define/namespace/option names", but otherwise LGTM.

@leopiccionia, yes, this PR does not attempt a full rename of every historical JEO identifier.

I checked the 2.15.1 package with Codex and most of the broader prefixing surface already existed there: the `Jeo` namespace, the global `Jeo` class, `JEO_*` constants, `jeo_*` helper functions, the `jeo-settings` option, the `map`, `map-layer`, and `storymap` post type slugs, and long-standing JS globals such as `jeo`, `jeo_settings`, `jeoMapVars`, `JeoLayerTypes`, `JeoLayer`, `JeoLegendTypes`, `JeoLegend`, `JeoInitMaps`, `jeomaps`, `initial_zoom`, `min_zoom`, `max_zoom`, and `layerFormData`.

The current PR only changes the newly reviewed low-risk runtime globals that were clearly generic in the discovery bundle: `languageParams` and `mapPreferences`, now `jeowpLanguageParams` and `jeowpMapPreferences`.

For the broader historical surface, I'm avoiding renaming them because it could mean other issues in terms of compatibility and migration. Some of those names are public integration points or stored identifiers.

Newer 3.0.0 additions are more limited: the `jeo/v1` route, new properties inside existing globals (`geocode_nonce`, `layersUrl`, map runtime fields), and `__jeoSetPanLimitsFromMap`. These either follow the existing JEO naming convention or live inside existing historical objects. If WordPress.org require a broader prefix migration, I'll handle it as a dedicated compatibility PR with migration notes and tests.

I will argue with them that these identifiers already existed and we probably shouldn't try to change them now unless they have a strong argument about that.

src/includes/class-jeo.php
bwstefano: Good point. I expanded this refactor slightly with commit [601b1ac](https://github.com/InfoAmazonia/jeo-plugin/pull/623/commits/601b1accffa1eb789f88c30a6af6557cde67ca43) because I found the same `disable_embed` decision was still duplicated inside the embed templates.

Since those templates are only included after the embed controller has resolved and validated the requested post, keeping access-control logic there would leave the behavior split between the controller and the view layer. The new helper centralizes the embed rendering guards for both maps and story maps, covering disabled embeds and password-protected posts in one place.

This keeps the behavior unchanged but makes the flow clearer: the controller decides whether an embed may render, and the templates only render already-authorized embed responses.

Issue #628, opened 2026-06-22
Replace Mapbox Static Tiles usage with per-map composed Mapbox styles


bwstefano:
## Background

Mapbox has asked us to optimize our API usage and reduce/remove usage of APIs such as the Static Tiles API. Their guidance was to move applications away from Static Tiles and toward the current Mapbox GL JS style rendering model. They also indicated that future discount/free-usage concessions depend on having a cost-efficient implementation.

The purpose of this issue is to track the implementation path we validated for JEO: replace the current Static Tiles rendering path for Mapbox style layers with a per-map composed Mapbox Style JSON, while preserving the current editorial model and runtime features.

## Current JEO behavior audited

JEO currently allows one map to be composed from multiple editorial layers. Those layers can be:

- `mapbox` layers, which point to complete Mapbox Studio styles.
- `mapbox-tileset-raster` layers.
- `mapbox-tileset-vector` layers.
- `mvt` layers.
- `tilelayer` raster tile URLs.

The current public map runtime chooses one `mapbox` layer as the base style when a layer is marked as `load_as_style`. Other Mapbox style layers are added as raster tiles from the Mapbox Static Tiles API. This was the workaround originally used to compose multiple Mapbox Studio styles over the same map.

The audited Static Tiles usage exists in at least these surfaces:

- Public map rendering: `src/includes/layer-types/mapbox.js` renders non-base `mapbox` layers with `https://api.mapbox.com/styles/v1/{styleId}/tiles/...`.
- Discovery: `src/js/src/discovery/blocks/map-layers.js` repeats a similar Static Tiles path for dynamic layer selection.
- Editor/preview rendering: `src/js/src/map-blocks/map-preview-layer.js` also renders `mapbox` layers as Static Tiles.

The public map runtime also contains important behavior that must be preserved:

- Layer order from `map.meta.layers`.
- `fixed`, `switchable`, and `swappable` layer use modes.
- Default visibility per layer.
- `style_layers` selection inside Mapbox styles.
- Layer legends.
- Layer attributions and source links.
- Related posts/markers.
- Click and mouse-over interactions configured in JEO.
- Storymaps derived from maps.
- Discovery maps.
- Adding and reordering layers in the editor.
- Support for MapLibre as the default rendering runtime.

## Agreed direction

Do not change the editorial model. Editors should continue to create maps, add layers, reorder layers, define default visibility, configure legends, and configure interactions as they do today.

[...]

PR #633, opened 2026-06-25
Replace Mapbox Static Tiles style overlays with composed Mapbox GL styles

bwstefano:
# Replace Mapbox Static Tiles style overlays with composed Mapbox GL styles

## Summary

This PR migrates JEO away from rendering `mapbox` layer posts as raster overlays through the Mapbox Static Tiles API. Instead, JEO now composes the Mapbox styles used by a map into a single Mapbox GL Style JSON and uses that composed style across regular maps, storymaps, Discovery, embeds, and editor previews.

The main goal is to preserve the current editorial model while removing the Static Tiles API usage generated by stacked Mapbox style layers. Editors can continue creating maps from multiple JEO layers, changing layer order, toggling default visibility, configuring switchable/swappable layers, adding legends, building storymaps from maps, using Discovery, and choosing Mapbox GL JS or MapLibre GL as the rendering library.

This change is also a response to Mapbox's request that InfoAmazonia reduce or eliminate Static Tiles API usage in order to keep the implementation cost-efficient and strengthen the case for future grant/discount renewals.

This work implements the migration mapped in #628.

## Why

Historically, JEO allowed editors to compose several Mapbox styles in one map by rendering one style as the WebGL base map and rendering additional Mapbox styles as raster tile overlays from the Static Tiles API. That solved an editorial need five years ago: different teams could maintain separate styles in Mapbox Studio and combine them in JEO without manually consolidating styles.

The downside is cost and platform alignment. Mapbox has specifically recommended moving applications away from APIs such as Static Tiles and toward current Mapbox GL JS usage. Their guidance ties a cost-efficient implementation to a stronger discount position in future cycles.

An analysis we made shows why this matters:

- In a representative high-composition map used for the estimate, the map had `9` editorial layers, `7` layers composed from Mapbox styles, `2` direct layers, `42` generated Mapbox GL style layers, and `31` generated style layers initially visible.
- In that scenario, the old model could generate roughly `48-120` Static Tiles API requests on a single initial map opening, because each additional Mapbox style layer was requested as a raster overlay.
- With the composer and Mapbox GL JS, the same composition moves to one Mapbox GL JS map load with vector/raster sources inside the GL style.
- InfoAmazonia's March-May 2026 Mapbox invoices show that Static Tiles API usage represented about `92.4%` of undiscounted Mapbox charges in those months.
- The three-month average Static Tiles line was about `USD 816.93/month`, or roughly `USD 9,803/year` annualized before discount.
- The composer does not target unrelated Mapbox costs such as tileset hosting or processing. Its target is specifically the Static Tiles API line generated by rendering extra Mapbox styles as raster overlays.

In short: this PR keeps the editorial flexibility that led JEO to use Static Tiles in the first place, but moves the rendering model to a modern Mapbox GL style composition approach.

### Backend composer

Adds a server-side Mapbox style composer that:

- Reads the map and layer configuration already stored in WordPress.
- Fetches referenced Mapbox styles.
- Namespaces source, layer, sprite, glyph, and metadata references to avoid collisions.
- Adds direct JEO layer types such as `mvt`, `tilelayer`, `mapbox-tileset-vector`, and `mapbox-tileset-raster`.
- Writes composed style artifacts and manifests to a cache directory.
- Supports temporary preview composition for editor and one-time map states that are not saved as a reusable map post.
- Preserves custom Mapbox tokens defined on individual layers.

### Shared frontend composer utilities

Adds shared JavaScript helpers for:

- Loading composed style metadata, style JSON, and manifests.
- Applying composed layer visibility.
- Finding composed manifest entries for original JEO layers.
- Wiring composed layer interactions and popups consistently across maps, storymaps, and Discovery.

### Regular maps and storymaps

Regular maps and storymaps now:

- Load the composed style when `mapbox` style layers are present.
- Use the composed manifest to preserve initial layer visibility and later layer toggles.
- Apply interactions from more than one Mapbox layer when the editor enables them.
- Show controlled warnings when composition is unavailable, instead of falling back to Static Tiles.
- Omit only problematic Mapbox style layers when needed, while continuing to render other direct layer types.

### Discovery and embeds

Discovery now uses composed Mapbox styles instead of Static Tiles when users combine maps/layers dynamically.

This includes:

- Loading composed map styles by map ID.
- Injecting composed sources and layers into the live Discovery map.
- Preloading sprite images needed by composed style layers.
- Preserving custom Mapbox access tokens from the selected layers.
- Keeping story pins above layer interaction popups.
- Adding layer interaction popups and map information popups with visual styling aligned with regular maps.
- Supporting exported Discovery embeds without storing an arbitrary user-generated style bundle on the server for every URL state.

### Editor and preview flows

Editor previews now use the composer too:

- Map post editor previews.
- Layer post editor previews.
- Storymap editor previews.
- One-time map block previews.
- Unsaved preview state through temporary composer payloads.

This means editors can see the current unsaved map/layer/storymap configuration before publishing, without relying on the removed Static Tiles path.

### Static Tiles removal

The legacy `mapbox` layer type no longer exposes an `addLayer()` implementation that creates a raster source from:

```text
https://api.mapbox.com/styles/v1/{styleId}/tiles/512/{z}/{x}/{y}@2x
```

`addStyle()`, `getStyleUrl()`, `getStyleLayers()`, and `addInteractions()` remain available for the code paths that still need style metadata or interaction definitions. Other layer types keep their existing rendering behavior.

This PR intentionally does not add a legacy Static Tiles fallback. If the composer cannot produce a valid style, JEO reports a controlled warning and avoids rendering only the affected Mapbox style layers.

### MapLibre compatibility

The composed style also works with MapLibre GL. The loader normalizes Mapbox style data where needed so MapLibre does not fail on unsupported or Mapbox-specific style properties.

Cost expectations differ by runtime:

- With Mapbox GL JS, the composed model aligns with Mapbox's map-load pricing model.
- With MapLibre GL, Mapbox-hosted vector/raster tiles may still be billed as tile requests, but the Static Tiles API path is still avoided.

## Editorial Behavior

The intended editorial workflow is mostly unchanged.

Editors still:

- Create reusable JEO maps from multiple layers.
- Add Mapbox style layers, MVT layers, tile layers, Mapbox vector tilesets, and Mapbox raster tilesets.
- Reorder layers.
- Decide which layers are active by default.
- Configure fixed, swappable, and switchable layers.
- Add legends and attribution.
- Create storymaps from maps.
- Add maps to Discovery.
- Export Discovery compositions as embeds.

The important conceptual change is internal: JEO now owns a style build/cache step. Editors do not need to manually merge styles in Mapbox Studio.

One visible editorial improvement is that `Load interactions` for Mapbox style layers is now an independent checkbox per layer. Previously the old raster-overlay model effectively required only one Mapbox layer to carry interactions, because other Mapbox styles were rendered as flat rasters. With the composer, interactions can be enabled for multiple composed Mapbox layers.

## Supporting Documentation

The implementation follows the transition specification in [mapbox-style-composition.md](https://github.com/user-attachments/files/29446536/mapbox-style-composition.md)

The cost and grant/discount rationale is documented in [mapbox-composer-cost-estimate.md](https://github.com/user-attachments/files/29446545/mapbox-composer-cost-estimate.md)

The cost estimate is especially relevant for review because this PR is not only a rendering refactor. It is also intended to materially reduce InfoAmazonia's Static Tiles API usage profile and support Mapbox discount/grant renewal conversations.

## Commits

This branch is organized into nine reviewable commits:

1. `Add Mapbox style composer backend`
2. `Add shared composed style frontend utilities`
3. `Render maps and storymaps with composed Mapbox styles`
4. `Remove legacy Static Tiles rendering for Mapbox layers`
5. `Use composed styles in editor map previews`
6. `Allow multiple Mapbox layers to load interactions`
7. `Use composed Mapbox styles in Discovery`
8. `Improve MapLibre handling for composed styles`
9. `Refresh translations and validation thresholds`

## Validation

Local validation performed during the branch:

- `vendor/bin/phpcs --standard=phpcs.xml.dist`
- `vendor/bin/phpcs --standard=phpcs-compat.xml.dist`
- `npm ci`
- `npm run build:assets`
- `npm run build`
- `npm run build:release`
- `npm run build:report`
- `npm run i18n:compile`
- WordPress smoke test through `scripts/wordpress-smoke.sh`
- Local WordPress Plugin Check run
- Manual checks in `infoamazonia.local` for maps, storymaps, Discovery, embeds, editor previews, Mapbox GL JS, and MapLibre GL

Additional static validation:

```bash
rg "/tiles/512|styles/v1/.*/tiles|Static Tiles|api\\.mapbox\\.com/styles/v1/.*/tiles" src
```

Expected result: no executable Static Tiles path remains in `src`.

## Notes From Manual Testing

The composer was validated visually against representative existing maps and storymaps in `infoamazonia.local`, including maps with:

- Multiple Mapbox style layers.
- `mapbox-tileset-raster` layers.
- `mapbox-tileset-vector` layers.
- MVT layers.
- Tile layers.
- Layer legends.
- Click and hover interactions.
- Storymap navigation.
- Discovery layer composition and iframe embeds.

The visual comparison was considered good enough to proceed with implementation. During testing we also confirmed that this change should not rely on theme-specific CSS fixes; theme problems should remain in the theme.

## Risks and Review Focus

Areas that deserve careful review:

- Style composition and cache invalidation in `Map_Style_Composer`.
- Sprite merging and image preloading, especially for styles with custom icons or patterns.
- Custom Mapbox token handling.
- Temporary composer payloads for unsaved editor previews and one-time maps.
- Discovery behavior when combining layers from multiple maps.
- Interaction popups when multiple composed Mapbox layers are active.
- MapLibre normalization of Mapbox style properties.
- Integration with future branches that still contain legacy Static Tiles code paths.

Important guardrail:

- Do not reintroduce `mapbox.addLayer()` as a raster Static Tiles renderer.
- Do not add a legacy Static Tiles fallback when composition fails.
- Do not add broad Mapbox v4 tile proxying as a fallback.

## Follow-up Work

Potential future improvements that are intentionally outside this PR:

- Add first-class interaction support for `mvt` and `mapbox-tileset-vector` layer types.
- Add UI-assisted source-layer/type discovery for vector tileset and MVT layers.
- Add simple color/opacity controls for supported vector layer types.
- Add deeper observability around composer cache hits, failures, and per-map composition diagnostics.
- Reconcile this branch carefully with `feature/experimental`, preserving this composer as the new baseline and removing the Static Tiles paths still present there.

### DIFF
--- a/src/includes/maps/class-map-style-composer.php
+++ b/src/includes/maps/class-map-style-composer.php
@@ -108,6 +108,21 @@ class Map_Style_Composer {
 
 		register_rest_route(
 			'jeo/v1',
+			'/map-style/layer/(?P<id>\d+)/refresh',
+			array(
+				'methods'             => WP_REST_Server::CREATABLE,
+				'callback'            => array( $this, 'refresh_layer_cache_response' ),
+				'permission_callback' => array( $this, 'can_refresh_layer_request' ),
+				'args'                => array(
+					'id' => array(
+						'sanitize_callback' => 'absint',
+					),
+				),
+			)
+		);
+
+		register_rest_route(
+			'jeo/v1',
 			'/map-style/(?P<scope>preview|onetime)/(?P<hash>[a-f0-9]{16})/style',
 			array(
 				'methods'             => WP_REST_Server::READABLE,
@@ -183,6 +198,21 @@ class Map_Style_Composer {
 	}
 
 	/**
+	 * Check whether the current request can force refresh a layer cache.
+	 *
+	 * @param WP_REST_Request $request REST request.
+	 * @return bool
+	 */
+	public function can_refresh_layer_request( WP_REST_Request $request ) {
+		$layer_id = absint( $request['id'] );
+		$post     = get_post( $layer_id );
+
+		return $post instanceof WP_Post &&
+			'map-layer' === $post->post_type &&
+			current_user_can( 'edit_post', $layer_id );
+	}
+
+	/**
 	 * Check whether the current request can read a virtual artifact.
 	 *
 	 * @param WP_REST_Request $request REST request.
@@ -205,9 +235,13 @@ class Map_Style_Composer {
 	 * @return WP_REST_Response
 	 */
 	public function get_metadata_response( WP_REST_Request $request ) {
+		$map_id        = absint( $request['id'] );
+		$force_refresh = rest_sanitize_boolean( $request->get_param( 'refresh' ) ) &&
+			current_user_can( 'edit_post', $map_id );
+
 		$result = $this->get_or_create_artifacts(
-			absint( $request['id'] ),
-			rest_sanitize_boolean( $request->get_param( 'refresh' ) )
+			$map_id,
+			$force_refresh
 		);
 
 		if ( is_wp_error( $result ) ) {
@@ -338,6 +372,49 @@ class Map_Style_Composer {
 	}
 
 	/**
+	 * Force refresh composed map style artifacts for maps using a layer.
+	 *
+	 * @param WP_REST_Request $request REST request.
+	 * @return WP_REST_Response
+	 */
+	public function refresh_layer_cache_response( WP_REST_Request $request ) {
+		$layer_id  = absint( $request['id'] );
+		$map_ids   = $this->get_map_ids_for_layer( $layer_id );
+		$results   = array();
+		$refreshed = 0;
+		$failed    = 0;
+
+		foreach ( $map_ids as $map_id ) {
+			$result = $this->get_or_create_artifacts( absint( $map_id ), true );
+
+			if ( is_wp_error( $result ) ) {
+				++$failed;
+				$results[] = array(
+					'mapId' => absint( $map_id ),
+					'error' => $result->get_error_message(),
+				);
+				continue;
+			}
+
+			++$refreshed;
+			$results[] = array(
+				'mapId' => absint( $map_id ),
+				'hash'  => $result['hash'] ?? null,
+			);
+		}
+
+		return rest_ensure_response(
+			array(
+				'layerId'   => $layer_id,
+				'mapIds'    => array_map( 'absint', $map_ids ),
+				'refreshed' => $refreshed,
+				'failed'    => $failed,
+				'results'   => $results,
+			)
+		);
+	}
+
+	/**
 	 * Clear map cache metadata when a map changes.
 	 *
 	 * @param int     $post_id Post ID.
@@ -363,6 +440,18 @@ class Map_Style_Composer {
 	public function invalidate_layer_cache( $post_id, $post, $update ) {
 		unset( $post, $update );
 
+		foreach ( $this->get_map_ids_for_layer( $post_id ) as $map_id ) {
+			$this->invalidate_map_cache( absint( $map_id ), null, true );
+		}
+	}
+
+	/**
+	 * Get map IDs that reference a layer in serialized map settings.
+	 *
+	 * @param int $layer_id Layer post ID.
+	 * @return int[]
+	 */
+	private function get_map_ids_for_layer( $layer_id ) {
 		$maps = get_posts(
 			array(
 				'post_type'      => 'map',
@@ -372,16 +461,14 @@ class Map_Style_Composer {
 				'meta_query'     => array(
 					array(
 						'key'     => 'layers',
-						'value'   => 's:2:"id";i:' . absint( $post_id ) . ';',
+						'value'   => 's:2:"id";i:' . absint( $layer_id ) . ';',
 						'compare' => 'LIKE',
 					),
 				),
 			)
 		);
 
-		foreach ( $maps as $map_id ) {
-			$this->invalidate_map_cache( absint( $map_id ), null, true );
-		}
+		return array_map( 'absint', $maps );
 	}
 
 	/**
--- a/src/includes/maps/class-map-style-composer.php
+++ b/src/includes/maps/class-map-style-composer.php
@@ -24,12 +24,12 @@ class Map_Style_Composer {
 
 	use Singleton;
 
-	const CACHE_DIR             = 'jeo-mapbox-composed-styles';
-	const CACHE_VERSION         = 11;
-	const TOKEN_PLACEHOLDER     = '__JEO_MAPBOX_ACCESS_TOKEN__';
-	const FALLBACK_SPRITE       = 'mapbox://sprites/mapbox/streets-v11';
-	const VIRTUAL_SCOPE_PREVIEW = 'preview';
-	const VIRTUAL_SCOPE_ONETIME = 'onetime';
+	const CACHE_DIR               = 'jeo-mapbox-composed-styles';
+	const CACHE_VERSION           = 11;
+	const TOKEN_PLACEHOLDER       = '__JEO_MAPBOX_ACCESS_TOKEN__';
+	const DEFAULT_FALLBACK_SPRITE = 'mapbox://sprites/mapbox/streets-v11';
+	const VIRTUAL_SCOPE_PREVIEW   = 'preview';
+	const VIRTUAL_SCOPE_ONETIME   = 'onetime';
 
 	/**
 	 * Register hooks.
@@ -1372,6 +1372,22 @@ class Map_Style_Composer {
 	}
 
 	/**
+	 * Return the fallback sprite root used for common missing Mapbox icons.
+	 *
+	 * Return an empty value from the filter to disable fallback sprite loading.
+	 *
+	 * @return string
+	 */
+	private function get_fallback_sprite() {
+		return trim(
+			(string) apply_filters(
+				'jeo_mapbox_composed_style_fallback_sprite',
+				self::DEFAULT_FALLBACK_SPRITE
+			)
+		);
+	}
+
+	/**
 	 * Merge sprites used by remote styles.
 	 *
 	 * @param array $bundles Style bundles.
@@ -1403,6 +1419,8 @@ class Map_Style_Composer {
 		if ( ! function_exists( 'imagecreatefromstring' ) ) {
 			return new WP_Error( 'jeo_mapbox_composer_gd_missing', __( 'The PHP GD extension is required to merge Mapbox sprites.', 'jeowp' ) );
 		}
+
+		$fallback_sprite = $this->get_fallback_sprite();
 
 		foreach ( array( 1, 2 ) as $ratio ) {
 			$entries        = array();
@@ -1440,8 +1458,8 @@ class Map_Style_Composer {
 				}
 
 				$missing = array_diff( $this->collect_style_image_names( $bundle['style'] ), array_keys( $fetched['json'] ) );
-				if ( ! empty( $missing ) && null === $fallback_cache ) {
-					$fallback_cache = $this->fetch_sprite( self::FALLBACK_SPRITE, $bundle['ref']['token'], $ratio );
+				if ( ! empty( $missing ) && null === $fallback_cache && '' !== $fallback_sprite ) {
+					$fallback_cache = $this->fetch_sprite( $fallback_sprite, $bundle['ref']['token'], $ratio );
 				}
 
 				if ( ! empty( $missing ) && is_array( $fallback_cache ) ) {
--- a/src/includes/maps/class-map-style-composer.php
+++ b/src/includes/maps/class-map-style-composer.php
@@ -372,7 +372,7 @@ class Map_Style_Composer {
 				'meta_query'     => array(
 					array(
 						'key'     => 'layers',
-						'value'   => (string) absint( $post_id ),
+						'value'   => 's:2:"id";i:' . absint( $post_id ) . ';',
 						'compare' => 'LIKE',
 					),
 				),
--- a/src/includes/maps/class-map-style-composer.php
+++ b/src/includes/maps/class-map-style-composer.php
@@ -1,0 +1,2414 @@
+<?php
+/**
+ * Compose Mapbox styles for saved maps.
+ *
+ * @package Jeo
+ */
+
+namespace Jeo;
+
+use WP_Error;
+use WP_Post;
+use WP_REST_Request;
+use WP_REST_Response;
+use WP_REST_Server;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+/**
+ * Build cached Mapbox Style JSON artifacts from JEO map/layer metadata.
+ */
+class Map_Style_Composer {
+
+	use Singleton;
+
+	const CACHE_DIR             = 'jeo-mapbox-composed-styles';
+	const CACHE_VERSION         = 11;
+	const TOKEN_PLACEHOLDER     = '__JEO_MAPBOX_ACCESS_TOKEN__';
+	const FALLBACK_SPRITE       = 'mapbox://sprites/mapbox/streets-v11';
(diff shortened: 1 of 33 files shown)


src/js/src/map-blocks/mapbox-style-preview.js
```diff
@@ -0,0 +1,570 @@
...
+
+	return useMemo( () => {
+		return ( url, resourceType ) => {
+			let nextUrl = String( url || '' );
+			let transformed = { url: nextUrl };
+
+			if ( nextUrl.includes( TOKEN_PLACEHOLDER ) ) {
+				nextUrl = nextUrl
+					.split( TOKEN_PLACEHOLDER )
+					.join( encodeURIComponent( mapboxToken || '' ) );
```
leopiccionia: [`String.prototype.replaceAll`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll) is widely available in modern browsers.
```suggestion
				nextUrl = nextUrl.replaceAll(
				    TOKEN_PLACEHOLDER,
					encodeURIComponent( mapboxToken || '' )
				);
```

src/js/src/map-blocks/mapbox-style-preview.js
```diff
@@ -0,0 +1,570 @@
...
+			setState( {
+				error: null,
+				isLoading: false,
+				style: null,
+				viewState: null,
+			} );
+			return undefined;
+		}
+
+		let didCancel = false;
```
leopiccionia: It's better handled with an [`AbortSignal`](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal#examples).

src/js/src/discovery/blocks/map-layers.js
```diff
@@ -4,14 +4,35 @@ import { __ } from '@wordpress/i18n';
...
+	mergeRecordsByIdOrder,
+} from '../../shared/rest-records';
+import { addComposedInteractions } from '../../shared/composed-style-layers';
 import LoadingSpinner from './loading-spinner';
 import MapItem from './map-item';
 import Search from './search';
 
 const MAPS_PER_PAGE = 20;
 const MAP_COLLECTION_FIELDS = 'id,title,excerpt,meta';
+const DEFAULT_GLYPHS = 'mapbox://fonts/mapbox/{fontstack}/{range}.pbf';
```
leopiccionia: It should be a WordPress filter, on PHP side.

src/js/src/discovery/index.js
```diff
@@ -220,6 +263,71 @@ class Discovery extends Component {
...
+	getInitialMapStyle() {
+		if ( ! defaultStyle || typeof defaultStyle !== 'object' ) {
+			return defaultStyle;
+		}
+
+		return {
+			...defaultStyle,
+			glyphs:
+				defaultStyle.glyphs ||
+				'mapbox://fonts/mapbox/{fontstack}/{range}.pbf',
```
leopiccionia: It should be a WordPress filter.

By the way, we shouldn't depend on it for MapLibreGL?

src/js/src/jeo-map/class-jeo-map.js
```diff
@@ -2059,6 +2212,13 @@ export default class JeoMap {
 	}
 
 	transformRequestUrl( url, resourceType ) {
+		const tokenPlaceholder = '__JEO_MAPBOX_ACCESS_TOKEN__';
+		if ( url.includes( tokenPlaceholder ) ) {
+			url = url.split( tokenPlaceholder ).join(
+				encodeURIComponent( jeo_settings.mapbox_key || '' )
+			);
```
leopiccionia:
```suggestion
			url = url.replaceAll(
			    tokenPlaceholder,
				encodeURIComponent( jeo_settings.mapbox_key || '' )
			);
```

src/includes/maps/class-map-style-composer.php
```diff
@@ -0,0 +1,2414 @@
...
+ * Build cached Mapbox Style JSON artifacts from JEO map/layer metadata.
+ */
+class Map_Style_Composer {
+
+	use Singleton;
+
+	const CACHE_DIR             = 'jeo-mapbox-composed-styles';
+	const CACHE_VERSION         = 11;
+	const TOKEN_PLACEHOLDER     = '__JEO_MAPBOX_ACCESS_TOKEN__';
+	const FALLBACK_SPRITE       = 'mapbox://sprites/mapbox/streets-v11';
```
leopiccionia: It should be a WordPress hook.

How does it work with MapLibreGL?

src/includes/maps/class-map-style-composer.php
```diff
@@ -0,0 +1,2414 @@
...
+	 */
+	private function ends_with( $haystack, $needle ) {
+		$haystack = (string) $haystack;
+		$needle   = (string) $needle;
+		if ( '' === $needle ) {
+			return true;
+		}
+		return substr( $haystack, -strlen( $needle ) ) === $needle;
+	}
+}
```
leopiccionia: PHP 8.0 supports [`str_starts_with`](https://www.php.net/manual/en/function.str-starts-with.php) and [`str_ends_with`](https://www.php.net/manual/en/function.str-ends-with.php).

src/includes/maps/class-map-style-composer.php
```diff
@@ -0,0 +1,2414 @@
...
+
+	/**
+	 * Register hooks.
+	 *
+	 * @return void
+	 */
+	protected function init() {
+		add_action( 'rest_api_init', array( $this, 'register_rest_routes' ) );
+		add_action( 'save_post_map', array( $this, 'invalidate_map_cache' ), 10, 3 );
+		add_action( 'save_post_map-layer', array( $this, 'invalidate_layer_cache' ), 10, 3 );
```
leopiccionia, line 42: It should be possible to invalidate a layer manually, from layers editor, if style changed in Mapbox but post didn't change necessarily.

src/includes/maps/class-map-style-composer.php
```diff
@@ -0,0 +1,2414 @@
...
+		$maps = get_posts(
+			array(
+				'post_type'      => 'map',
+				'post_status'    => 'any',
+				'posts_per_page' => -1,
+				'fields'         => 'ids',
+				'meta_query'     => array(
+					array(
+						'key'     => 'layers',
+						'value'   => (string) absint( $post_id ),
```
leopiccionia: It's inherently buggy. Layer `234` or `123` shouldn't be invalidated because layer `23` was.

The following is a better match (see [PHP serialization format](https://en.wikipedia.org/wiki/PHP_serialization_format)):

```suggestion
						'value'   => 's:2:"id";i:' . absint( $post_id ) . ';',
```

src/js/src/jeo-map/class-jeo-map.js
```diff
@@ -205,70 +216,27 @@ export default class JeoMap {
...
-									} else {
-										// If the current layer is below the style, add using first syle layer reference
-										if ( index < styleLayerIndex ) {
-											layer.addLayer( map, [ firstStyleLayerId ] );
-										} else {
-											layer.addLayer( map );
-										}
+						this.mapLoaded.then( () => {
+							if ( usingComposedStyle ) {
+								addComposedInteractions( map, this.composedManifest );
```
leopiccionia: Cleanup functions are returned by `addComposedInteractions`, but never stored or called.

(Found by AI.)

src/includes/maps/class-map-style-composer.php
leopiccionia: By the way, it seems like `streets-v11` (and its `v12` counterpart) [is deprecated](https://docs.mapbox.com/map-styles/reference/streets/). Frowm what I understand, we are supposed to use whatever comes with [`mapbox/standard`](https://docs.mapbox.com/map-styles/reference/standard/).

src/js/src/map-blocks/mapbox-style-preview.js
bwstefano: Done in ac1beac6. The editor preview now uses `AbortController` and threads the signal through composed style metadata/style/manifest and TileJSON requests, while keeping the existing stale-state guard.

src/js/src/discovery/blocks/map-layers.js
bwstefano: Done in 2d50e393. The default glyph endpoint now comes from PHP via `jeo_mapbox_composed_style_default_glyphs` and is localized to the frontend instead of being hardcoded in Discovery.

src/js/src/discovery/index.js
bwstefano: Done in 2d50e393. The Discovery fallback glyph endpoint is now PHP-filtered, and an empty filtered value disables the fallback. MapLibre-first deployments can provide their own glyph endpoint instead of inheriting the Mapbox default.

src/includes/maps/class-map-style-composer.php
bwstefano: Done in 2d50e393. The composer fallback sprite now goes through `jeo_mapbox_composed_style_fallback_sprite`, and returning an empty value disables fallback sprite loading. This keeps the composed style runtime-neutral while allowing MapLibre-first deployments to avoid an implicit Mapbox sprite fallback.

src/includes/maps/class-map-style-composer.php
bwstefano: Done in d068cff8. The composer now uses native `str_starts_with()` and `str_ends_with()` directly, and the local wrappers were removed.

src/includes/maps/class-map-style-composer.php
bwstefano: Done in 2734179c. The layer editor now has an authenticated refresh action for Mapbox layers. It refreshes composed artifacts for maps that reference the layer, and public map reads no longer honor `refresh=1` unless the current user can edit the map.

src/js/src/jeo-map/class-jeo-map.js
bwstefano: Done in e8916b00. The regular map and storymap runtimes now store and run the cleanup callbacks returned by `addComposedInteractions` before re-registering interactions and on teardown.

bwstefano:
> By the way, it seems like `streets-v11` (and its `v12` counterpart) [is deprecated](https://docs.mapbox.com/map-styles/reference/streets/). Frowm what I understand, we are supposed to use whatever comes with [`mapbox/standard`](https://docs.mapbox.com/map-styles/reference/standard/).

@leopiccionia, this is also done in db3c5eb8 and 7149d5f9. The composer fallback sprite now uses `mapbox://sprites/mapbox/standard` with a cache-version bump, and the Mapbox GL default style now uses `mapbox://styles/mapbox/standard`. The fallback remains filterable.

src/js/src/map-blocks/mapbox-style-preview.js
leopiccionia: Can we remove the `didCancel` at this point?

bwstefano: Done in f86b7572. I removed the separate `didCancel` flag and now use the `AbortSignal` itself as the stale-request guard, while still skipping state updates for aborted preview requests.

leopiccionia:
Esse PR, em tese, cria dois tipos diferentes de `layer-types`: os tipos de estilo (atualmente apenas `mapbox`) e os demais.

Acho que a forma como o tipo `mapbox` é tratado de forma especial é um débito técnico, mas pode ser resolvido numa versão futura (criando uma API clara que separe as instâncias dos dois tipos de camadas).

De resto, parece estar num estado bom para ser mergeado e testado.

leopiccionia (approved):

